PR #67, opened 2017-07-05
Minor API changes to get the signal single point example working

jashnani:
- [ ] `tox` successfully runs, including unit tests and style checks (see [CONTRIBUTING.md](https://github.com/ni/nixnet-python/blob/master/CONTRIBUTING.rst)).

Minor API changes for getting the CAN signal single point example working.

Gets one of the examples working.

Successfully ran the example on a test machine with driver installed.

### DIFF
--- a/nixnet/_ctypedefs.py
+++ b/nixnet/_ctypedefs.py
@@ -105,11 +105,11 @@ class nxDatabaseRef_t(u32):  # NOQA: N801
     pass
 
 
-class nxStatus_t(i32):  # NOQA: N801
+class nxStatus_t(u32):  # NOQA: N801
     pass
 
 
-class nxTimeStamp_t(u64):  # NOQA: N801
+class nxTimestamp_t(u64):  # NOQA: N801
     """"Absolute time, given in 100 ns increments since Jan 1, 1601, 12:00 AM UTC."""
     pass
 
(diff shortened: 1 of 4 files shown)


nixnet/_ctypedefs.py
```diff
@@ -105,7 +105,7 @@ class nxDatabaseRef_t(u32):  # NOQA: N801
     pass
 
 
-class nxStatus_t(i32):  # NOQA: N801
+class nxStatus_t(u32):  # NOQA: N801
```
epage: Please document the reason for this in your commit message.

Well, since its more of a "I told you to", I guess I'll give a write-up
> Change `nxStatus_t` to be a `u32` instead of `i32`.  The problem is our constants are defined as unsigned while `nxStatus_t` was defined as signed.  This caused several problems (incorrect detection of errors, failure to convert code to enum value).
> For now, we are going with the "quickest to get working" solution.  In the long term, we need to decide if we should treat these as unsigned or signed.

nixnet_examples/can_signal_single_point_io.py
```diff
@@ -62,14 +74,14 @@ def main():
                 num_signals = len(value_buffer)
                 signals = input_session.read_signal_single_point(num_signals)
                 for timestamp, value in signals:
-                    date = datetime.datetime.fromtimestamp(timestamp / 1e9)
+                    date = datetime.datetime.fromtimestamp(timestamp / 1e9) + delta
```
epage: I assume this is still not working.

its fine if it doesn't, I just wanted to confirm

jashnani: Yes, the timestamp code is still not working

nixnet/_ctypedefs.py
jashnani: Added it to the commit message

Issue #23, opened 2017-06-16
Design OOP API for vertical slice


epage:
Refactor the skeleton API to be OOP
- [x] #22 API is sufficient for examples
- [x] Basic session properties
- [x] Session sub-properties
- [x] Evaluate adding hierarchy to the properties
- [x] Read/write API

epage:
For the read/write API, the thing I'm considering

We define a BytesBuffer and FloatBuffer.  Reading raw frame and reading of signals will use these and return the values contained in them.  Compared to these `read` functions, we'll have `read_into` functions that take in the buffers.  We then manipulate them as needed for the number of samples requested to use in our calls.  This allows buffer reuse for performance.

If possible, these buffer objects would mimic enough of numpy that read/write need for numpy to be a drop in replacement, if needed, without having a hard dependency on numpy.

epage:
Additionally for read/write, unlike DAQmx, XNET works more like [`io`](https://docs.python.org/3.5/library/io.html#class-hierarchy), should we implement one of those interfaces?

XNET-PSE:
I agree XNET works very much like a stream. Many of the interfaces are already implemented in the C api. Read, write, flush, etc. Implementing them in the Python API seems like a good fit.

PR #80, opened 2017-07-10
feat(API): Split `Session` into classes per-mode

epage:
This is a part of #23 where we are trying to get the API stable-ish.

Benefits
- Guide the user to relevant API functionality rather than burying it in
  docs or waiting until runtime errors
- Tailored `Session.__init__` for the mode, like LabVIEW
- Static design (read/write don't dynamically adapt based on the mode),
  allowing for improved static analysis.

Downsides
- Reads and writes are common operations but buried one level deeper.
  - But similar to DAQmx's `stream` object but more context aware (but
    DAQmx has "simple read/write" flavors)
  - `Session.frames` and `Session.signals` already existed so it can
    logically make sense to put these operation on those objects.
  - The reads/writes would already have "signals" and "frames" in there
    name, so the overall length isn't any longer.
  - Less code duplication because a `Frames` object can be reused across
    modes.

Concerns
- Number of modes this will lead to once we've implemented them all
- There might be properties that do not apply to all modes, especially the
  conversion modes.
- That there might be some more mode-specific semantics we want to expose

### Boilerplate

None

### DIFF
--- a/nixnet/__init__.py
+++ b/nixnet/__init__.py
@@ -6,6 +6,6 @@ from __future__ import unicode_literals
 from nixnet.errors import XnetError  # NOQA: F401
 from nixnet.errors import XnetResourceWarning  # NOQA: F401
 from nixnet.errors import XnetWarning  # NOQA: F401
-from nixnet.session import Session  # NOQA: F401
+from nixnet.session import *  # NOQA: F401, F403
 
 __all__ = []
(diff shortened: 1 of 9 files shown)


epage:
Just got confirmation from @XNET-PSE  that single point reads should not ask for the number of samples

PR #83, opened 2017-07-11
feat(tests): Example of integration tests

epage:
This is leveraging py.test for both unit and integration tests.

Hardware is auto-selected but can be overridden with an environment
variable.  A `fixture` was used to pass this to the test (rather than the
test reading from `request` directly) so that if the test fails, pytest
will show what hardware fixture was used.

- [ ] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nixnet-python/blob/master/CONTRIBUTING.rst).
- [ ] `tox` successfully runs, including unit tests and style checks (see [CONTRIBUTING.md](https://github.com/ni/nixnet-python/blob/master/CONTRIBUTING.rst)).

TODO: Check the above boxes with an 'x' to indicate what steps you have taken in preparing this Pull Request.

TODO: Include high-level description of the changes in this pull request.

TODO: Justify why this contribution should be part of the project.

TODO: Detail what testing has been done to ensure this submission meets requirements.


epage:
This is meant as an example and not meant to be directly merged.

The main problem I'm running into at the moment (besides the test not working) is that `integration` only runs with a single python version rather than across all of them.

epage:
So we use tox-travis to automatically connect the travis matrix to the tox matrix.  The problem is that that will automatically run the integration tests.

I tried to work around this by overriding the envlist within travis but that seems to mess up tox-travis
```
    - TOXENV=py{,2,27,3,34,35,36,py,py3}-test,flake8,docs,pkg
```

tests/test_signals.py
```diff
@@ -0,0 +1,89 @@
...
+            output_session.signals.write(expected_signals)
+
+            # Wait 1 s and then read the received values.
+            # They should be the same as the ones sent.
+            time.sleep(1)
+
+            actual_signals = list(input_session.signals.read())
+            for expected, (_, actual) in zip(expected_signals, actual_signals):
+                assert pytest.approx(expected, rel=1) == actual
+
```
lilesj: We should take care when comparing the expected results from signals. They start as doubles, get scaled, offset, and ultimately converted to integers and the process is reversed on the receiving end. Sometimes the floating point math can introduce anomalies where 17.0 ends up being received as 16.9999999 or something similar.

epage: That is what `pytest.approx` is for.  I'm using a relative comparison to try to compensate for it.

lilesj: Perfect. Looks good to me then.

PR #97, opened 2017-07-17
Adding RST files + Documentation cleanup

jashnani:

-Adds reStructuredText files for Sphinx documentation generation support. 
-Switches to using `:any:` for cross-referencing as Sphinx generates warnings for bad targets (helps catch broken links).
-Fixes broken references for functions/classes
-Other minor changes/fixes for Sphinx to work

Note that doc references for enums, constants and properties are still broken, will address this in a separate PR.

Adds documentation generation support for the 0.1 release.

Ran Sphinx and reviewed the generated documentation.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -12,6 +12,18 @@ NI-XNET Python Documentation
    :maxdepth: 3
    :caption: API Reference:
 
+   session
+   frames
+   signals
+   intf
+   j1939
+   types
+   errors
+
+   base
+   enums
+   collection
+
 
 Indices and Tables
 ==================
--- a/nixnet/errors.py
+++ b/nixnet/errors.py
@@ -21,8 +21,8 @@ class XnetError(Error):
         """Initialize error.
 
         Args:
-            message (string): Specifies the error message.
-            error_code (int): Specifies the NI-XNET error code.
+            message: A string specifing the error message.
+            error_code: An integer specifing the NI-XNET error code.
         """
         super(XnetError, self).__init__(message)
 
@@ -37,7 +37,8 @@ class XnetError(Error):
     def error_code(self):
         """Error code reported by NI-XNET.
 
-        int: Specifies the NI-XNET error code.
+        Returns:
+            An integer specifing the NI-XNET error code.
         """
         return self._error_code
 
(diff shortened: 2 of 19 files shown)


epage:
So you can't write `__doc__` to a namedtuple.

See https://stackoverflow.com/questions/1606436/adding-docstrings-to-namedtuples

nixnet/errors.py
```diff
@@ -21,8 +21,8 @@ def __init__(self, message, error_code):
         """Initialize error.
 
         Args:
-            message (string): Specifies the error message.
-            error_code (int): Specifies the NI-XNET error code.
+            message: A string specifing the error message.
+            error_code: An integer specifing the NI-XNET error code.
```
epage: So google's docstring style isn't very well documented.  I found some examples of it though
http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

The former is mostly correct (`str` instead of `string`)

nixnet/errors.py
```diff
@@ -74,16 +75,17 @@ def __init__(self, message, error_code):
     def error_code(self):
         """Error code reported by NI-XNET.
 
-        int: Specifies the NI-XNET error code.
+        Returns:
```
epage: From the example I found, looks like we are both wrong on how to document a property

docs/index.rst
```diff
@@ -12,6 +12,17 @@ NI-XNET Python Documentation
...
+   session
+   base
+   frames
+   signals
+   types
+   errors
+   constants
+   collection
+   intf
+   j1939
```
epage: [DAQmx only included top level modules](https://raw.githubusercontent.com/ni/nidaqmx-python/master/docs/index.rst) and left things like the following to the sub pages
- frames
- signals
- intf
- j1939

My hope is we can get away without directly listing
- base
- collection

nixnet/errors.py
jashnani: This will be a big change so I'll do it in a separate PR. For now I've modified types, errors, etc. to match majority of the docstrings in the project.

nixnet/errors.py
jashnani: Ditto as above.

docs/index.rst
jashnani: We have references to things in 'base.py' for which Sphinx complains if an RST file is created but not added to a TOCtree. Maybe there's a way to have it be generat doc content without adding to a TOC?

We need to revamp 'index.rst' anyway (because it's only showing API reference in the TOC section) so we can fix this then.

We also need to cleanup the code for relying documentaion in '_enums.py', i.e. enums.rst.
Ideally that documentation would go in a public module....maybe 'constants.py' but I didn't have the time to do that work.

Issue #127, opened 2017-07-21
Exception when loading DLL


ellinorcrux:
We upgraded Python, but now we're getting a different Exception: 

```
    self._instream = FrameInStreamSession(self._interface_name, self._database_n
ame, self._cluster_name)
  File "C:\Python27\lib\site-packages\nixnet\session.py", line 86, in __init__
    constants.CreateSessionMode.FRAME_IN_STREAM)
  File "C:\Python27\lib\site-packages\nixnet\_session\base.py", line 59, in __in
it__
    self._handle = _funcs.nx_create_session(database_name, cluster_name, list, i
nterface_name, mode)
  File "C:\Python27\lib\site-packages\nixnet\_funcs.py", line 32, in nx_create_s
ession
    ctypes.pointer(session_ref_ctypes),
  File "C:\Python27\lib\site-packages\nixnet\_cfuncs.py", line 37, in nx_create_
session
    cfunc = self.cdll.nxCreateSession
  File "C:\Python27\lib\site-packages\nixnet\_cfuncs.py", line 26, in cdll
    self._cdll = _lib.import_lib()
  File "C:\Python27\lib\site-packages\nixnet\_lib.py", line 56, in _import_win_l
ib
    cdll = ctypes.cdll.LoadLibrary(lib_name)
  File "C:\Python27\lib\ctypes\__init__.py", line 440, in LoadLibrary
    return self._dlltype(name)
  File "C:\Python27\lib\ctypes\__init__.py", line 362, in __init__
    self._handle = _dlopen(self._name, mode)
TypeError: LoadLibrary() argument 1 must be string, not unicode
```

- OS: Windows 7 
- Python version: Python 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:42:59)
- NI-XNET version:  Driver version 16,  nixnet (0.1.0.dev2)

epage:
Apparently, this is a [bug](https://bugs.python.org/issue29082) in Python 2.7.13 that should be fixed in 2.7.14 which should be released [any day now](https://www.python.org/dev/peps/pep-0373/).

It looks like DAQmx also ran into this, see ni/nidaqmx#1.

Would you prefer us to hack around this or to downgrade?

ellinorcrux:
[...]

Issue #116, opened 2017-07-20
Exception on import of _frames


ellinorcrux:
Getting an Exception on import from _frames module without a helpful error message. 

## traceback: 

```
  File "ni_xnet_transport.py", line 7, in <module>
    from nixnet import FrameInStreamSession, FrameOutStreamSession
  File "C:\Python27\lib\site-packages\nixnet\__init__.py", line 9, in <module>
    from nixnet.session import *  # NOQA: F401, F403
  File "C:\Python27\lib\site-packages\nixnet\session.py", line 11, in <module>
    from nixnet._session import frames as session_frames
  File "C:\Python27\lib\site-packages\nixnet\_session\frames.py", line 8, in <mo
dule>
    from nixnet import _frames
  File "C:\Python27\lib\site-packages\nixnet\_frames.py", line 15, in <module>
    nxFrameFixed_t = struct.Struct('QIBBBB8s')  # NOQA: N801
TypeError: Struct() argument 1 must be string, not unicode
```

- OS: Windows 7
- Python version: 2.7 (r27:82525, Jul  4 2010, 09:01:59) [MSC v.1500 32 bit (Intel)]
- NI-XNET version: nixnet (0.1.0.dev2) (the one from today), XNET drivers: version 16

### Installed packages and version
backports-abc (0.5)
backports.ssl-match-hostname (3.5.0.1)
certifi (2017.4.17)
chardet (3.0.4)
click (6.7)
enum34 (1.1.6)
Flask (0.12.2)
future (0.16.0)
fuzzywuzzy (0.15.1)
hoplite (15.0.0.dev21)
idna (2.5)
itsdangerous (0.24)
Jinja2 (2.9.6)
labview-automation (15.0.0.dev23)
labview-io (0.0.4)
MarkupSafe (1.0)
nidaqmx (0.5.2)
nifpga (17.0.0.post2)
nixnet (0.1.0.dev2)
numpy (1.13.0)
pip (9.0.1)
psutil (5.2.2)
PyDAQmx (1.3.2)
pymongo (3.4.0)
pyserial (3.3)
PyVISA (1.8)
pywin32 (221)
requests (2.18.1)
robotframework (3.0.2)
setuptools (25.2.0)
singledispatch (3.4.0.3)
six (1.10.0)

epage:
Hey, thanks for letting us know about this!

My suspicion is that this was an issue fixed between version of Python 2.7.  For example
```
> python
[...]

PR #130, opened 2017-07-24
Adding environment variables to Jenkinsfile

lilesj:

Established environment variables in the Jenkinsfile to specify hardware for CI testing.

To enable CI testing with hardware.

Verified Jenkins executes test script on my dev fork.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,18 +1,37 @@
 #!groovy
-node('xnetPython01') {
+node('xnetPython') {
 
 	currentBuild.result = "SUCCESS"
 	
-	try{
-		stage('Test'){
-			// test script to checkout the scm
-			checkout scm
-		}		
+	// Environment variables to configure the hardware for testing. This scope applies to all stages
+	environment {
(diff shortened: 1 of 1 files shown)


epage:
py2-test failed because of #116.  I'm working on a fix for that but it does raise the question, why are we running the tests on Python 2.7.0 instead of the latest Python 2.7 release?

Granted, the latest python release is also broken, see #127.  My fix will hopefully fix both.

lilesj:
I installed 2.7.0 because it is the earliest version that we claimed to support at the time of setup. It looks like we have since modified our claimed support to 2.7.7 so I can update the test machine accordingly.

epage:
Previously, I was only aware of us claiming 2.7.  I guess the question is does that presume the latest or earliest 2.7.

epage:
I'm confirming with our other python projects for consistency purposes.

When we decided on Python 2.7 support, I presume Python 2.7.<latest> because those dot-releases are bug fixes and we should only guarantee support with all the latest bug fixes, similar to when an SP1 of LabVIEW comes out, we stop supporting the non-SP1 (I think).

PR #131, opened 2017-07-24
feat(state): Report general and CAN session state

epage:
There are two controversial parts of this

- If the interface isn't communication, time_start and time_communicating
  throw an error rather than return 0.
- Each state property will treat `fault` like any other return code,
  raising an exception or warning, when these properties are accessed

Standard checklist


epage:
Originally, each state property would call `check_for_error` on the fault.  After talking with Christian, I switched to an explicit `check_fault` function.

PR #142, opened 2017-07-25
feat(API): Support lin_sched

bcornish:
- [ ]  Document example
- [x]  Integration tests: refactor can frame parameters
- [x]  Add integration tests: lin frame

Added read and write frame for LIN. Added LIN frame object.

Stepping stone for LIN support.

Doc tests only

### DIFF
--- a/nixnet/_session/base.py
+++ b/nixnet/_session/base.py
@@ -336,7 +336,7 @@ class SessionBase(object):
         _funcs.nx_disconnect_terminals(self._handle, source, destination)
 
     def change_lin_schedule(self, sched_index):
-        # type: (typing.Text) -> None
+        # type: (int) -> None
         """Writes communication states of an XNET session.
 
         This function writes a request for the LIN interface to change
--- a/nixnet/_session/base.py
+++ b/nixnet/_session/base.py
@@ -335,6 +335,41 @@ class SessionBase(object):
         """
         _funcs.nx_disconnect_terminals(self._handle, source, destination)
 
+    def change_lin_schedule(self, sched_index):
+        # type: (typing.Text) -> None
+        """Writes communication states of an XNET session.
+
+        This function writes a request for the LIN interface to change
+        the running schedule.
+
+        According to the LIN protocol, only the master executes schedules,
+        not slaves. If the
+        :any:`nixnet._session.intf.Interface.lin_master` property is false (slave),
+        this write function implicitly sets that property to true (master). If the
+        interface currently is running as a slave, this write returns an error,
+        because it cannot change to master while running.
+
+        Args:
+            sched_index(int): Index to the schedule table that the LIN master executes.
+
+            The schedule tables are sorted the way they are returned from the database
+            with the `nixnet.db._cluster.Cluster.lin_schedules` property.
+        """
+        _funcs.nx_write_state(self._handle, constants.WriteState.LIN_SCHEDULE_CHANGE, _ctypedefs.u32(sched_index))
+
+    def change_lin_diagnostic_schedule(self, schedule):
+        # type: (constants.LinDiagnosticSchedule) -> None
+        """Writes communication states of an XNET session.
+
+        This function writes a request for the LIN interface to change
+        the diagnostic schedule.
+
+        Args:
+            schedule(:any:`nixnet._enums.LinDiagnosticSchedule`): Diagnostic schedule
+                that the LIN master executes.
+        """
+        _funcs.nx_write_state(self._handle, constants.WriteState.LIN_DIAGNOSTIC_SCHEDULE_CHANGE, _ctypedefs.u32(schedule.value))  # NOQA: E501
+
     @property
     def time_current(self):
         # type: () -> int
--- a/nixnet/_session/intf.py
+++ b/nixnet/_session/intf.py
@@ -776,6 +776,26 @@ class Interface(object):
 
     @property
     def lin_master(self):
+        '''boolean: LIN Master?
+
+        Specifies the NI-XNET LIN interface role on the network: master (true)
+        or slave (false).
+
+        In a LIN network (cluster), there always is a single ECU in the system
+        called the master. The master transmits a schedule of frame headers.
+        Each frame header is a remote request for a specific frame ID. For each
+        header, typically a single ECU in the network (slave) responds by transmitting
+        the requested ID payload. The master ECU can respond to a specific header
+        as well, and thus the master can transmit payload data for the slave
+        ECUs to receive.
+
+        The default value for this property is false (slave). This means that by
+        default, the interface does not transmit frame headers onto the network.
+        When you use input sessions, you read frames that other ECUs transmit.
+        When you use output sessions, the NI-XNET interface waits for the remote
+        master to send a header for a frame in the output sessions, then the interface
+        responds with data for the requested frame.
+        '''
         return _props.get_session_intf_lin_master(self._handle)
 
     @lin_master.setter
--- a/nixnet/_session/intf.py
+++ b/nixnet/_session/intf.py
@@ -776,6 +776,7 @@ class Interface(object):
 
     @property
     def lin_master(self):
+        # type: () -> bool
         '''boolean: LIN Master?
 
         Specifies the NI-XNET LIN interface role on the network: master (true)
--- a/tox-integration.ini
+++ b/tox-integration.ini
@@ -9,12 +9,14 @@ skip_missing_interpreters=True
 [testenv]
 description = Run tests
 commands =
-    test: pytest --junit-xml=junit/{envname}.xml --junit-prefix={envname} {posargs: --can-in-interface {env:CAN_FIXTURE_IN_INTERFACE:CAN1} --can-out-interface {env:CAN_FIXTURE_OUT_INTERFACE:CAN2}}
+    test: pytest --junit-xml=junit/{envname}.xml --junit-prefix={envname} {posargs: --can-in-interface {env:CAN_FIXTURE_IN_INTERFACE:CAN1} --can-out-interface {env:CAN_FIXTURE_OUT_INTERFACE:CAN2} --lin-in-interface {env:LIN_FIXTURE_IN_INTERFACE:LIN1} --lin-out-interface {env:LIN_FIXTURE_OUT_INTERFACE:LIN2}}
 deps =
     -rrequirements_test.txt
 passenv =
     CAN_FIXTURE_IN_INTERFACE
     CAN_FIXTURE_OUT_INTERFACE
+    LIN_FIXTURE_IN_INTERFACE
+    LIN_FIXTURE_OUT_INTERFACE
 
 [pytest]
 addopts = --cov nixnet --cov nixnet_examples --cov-report term --cov-report xml --verbose --doctest-modules --ignore=setup.py
--- a/tox-integration.ini
+++ b/tox-integration.ini
@@ -9,12 +9,12 @@ skip_missing_interpreters=True
 [testenv]
 description = Run tests
 commands =
-    test: pytest --junit-xml=junit/{envname}.xml --junit-prefix={envname} {posargs: --nixnet-in-interface {env:NIXNET_FIXTURE_IN_INTERFACE:CAN1} --nixnet-out-interface {env:NIXNET_FIXTURE_OUT_INTERFACE:CAN2}}
+    test: pytest --junit-xml=junit/{envname}.xml --junit-prefix={envname} {posargs: --can-in-interface {env:CAN_FIXTURE_IN_INTERFACE:CAN1} --can-out-interface {env:CAN_FIXTURE_OUT_INTERFACE:CAN2}}
 deps =
     -rrequirements_test.txt
 passenv =
-    NIXNET_FIXTURE_IN_INTERFACE
-    NIXNET_FIXTURE_OUT_INTERFACE
+    CAN_FIXTURE_IN_INTERFACE
+    CAN_FIXTURE_OUT_INTERFACE
 
 [pytest]
 addopts = --cov nixnet --cov nixnet_examples --cov-report term --cov-report xml --verbose --doctest-modules --ignore=setup.py
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,6 +7,8 @@ node('xnetPython') {
 	environment {
 		CAN_FIXTURE_IN_INTERFACE = 'CAN1'
 		CAN_FIXTURE_OUT_INTERFACE = 'CAN2'
+		LIN_FIXTURE_IN_INTERFACE = 'LIN1'
+		LIN_FIXTURE_OUT_INTERFACE = 'LIN2'
 	}
 
 	try{
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,8 +5,8 @@ node('xnetPython') {
 
 	// Environment variables to configure the hardware for testing. This scope applies to all stages
 	environment {
-		NIXNET_FIXTURE_IN_INTERFACE = 'CAN1'
-		NIXNET_FIXTURE_OUT_INTERFACE = 'CAN2'
+		CAN_FIXTURE_IN_INTERFACE = 'CAN1'
+		CAN_FIXTURE_OUT_INTERFACE = 'CAN2'
 	}
 
 	try{
--- a/nixnet/_enums.py
+++ b/nixnet/_enums.py
@@ -1406,6 +1406,62 @@ class LinCommState(enum.Enum):
 
 
 class LinDiagnosticSchedule(enum.Enum):
+    """LIN Diagnostic Schedule
+
+    Values:
+        NULL:
+            The master does not execute any diagnostic schedule. No master
+            request or slave response headers are transmitted on the LIN.
+        MASTER_REQ:
+            The master executes a diagnostic master request schedule
+            (transmits a master request header onto the LIN) if it can.
+            First, a master request schedule must be defined for the LIN
+            cluster in the imported or in-memory database. Otherwise, error
+            'nixnet._enums.Err.DIAGNOSTIC_SCHEDULE_NOT_DEFINED' is returned
+            when attempting to set this value. Second, the master must have
+            a frame output queued session created for the master request frame,
+            and there must be one or more new master request frames pending in
+            the queue. If no new frames are pending in the output queue, no
+            master request header is transmitted. This allows the timing of
+            master request header transmission to be controlled by the timing
+            of master request frame writes to the output queue.
+
+            If there are no normal schedules pending, the master is effectively
+            in diagnostics-only mode, and master request headers are transmitted
+            at a rate determined by the slot delay defined for the master request
+            frame slot in the master request schedule or the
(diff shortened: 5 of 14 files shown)


nixnet/_session/frames.py
```diff
@@ -159,6 +159,34 @@ def read_can(
         for frame in self.read_raw(number_to_read, timeout):
             yield types.CanFrame.from_raw(frame)
 
+    def read_lin(
+            self,
+            number_to_read,
+            timeout=constants.TIMEOUT_NONE):
```
epage: Please apply a type annotation

I'm guessing it'd be
```python
# type: (int, float) -> typing.Iterable[types.LinFrame]
```

epage: ditto for all other functions

nixnet/types.py
```diff
@@ -220,3 +221,110 @@ def __repr__(self):
...
+    Attributes:
+        identifier: An integer representing the LIN frame arbitration identifier.
+        echo: A boolean indicating if the frame is an echo of a successful
+            transmit rather than being received from the network.
+        type: Frame type as described in :any:`nixnet._enums.FrameType`.
+        timestamp: Absolute time the XNET interface received the end-of-frame.
+        payload: A byte string representing the payload.
+        eventid: Identifier for an event-triggered slot.
+        eventslot: A boolean indicating whether the frame was received within an
+            event-triggered slot or an unconditional or sporadic slot.
```
epage: Please document the expected type for each of these

nixnet/types.py
```diff
@@ -220,3 +221,110 @@ def __repr__(self):
...
+        "echo",
+        "type",
+        "timestamp",
+        "eventid",
+        "eventslot",
+        "payload"]
+
+    _FRAME_ID_MASK = 0x000000FF
+
+    def __init__(self, identifier, type, eventid=b"", payload=b""):
```
epage: why do you default `eventid` to a bytestring here but pass in an int elsewhere?  What should its type be?

(also, this is the failure that Travis caught)

nixnet/types.py
```diff
@@ -220,3 +221,110 @@ def __repr__(self):
...
+    def __repr__(self):
+        # type: () -> typing.Text
+        """LinFrame debug representation.
+
+        >>> LinFrame(2, constants.FrameType.LIN_DATA, 0)
+        LinFrame(identifier=0x2, echo=False, type=FrameType.LIN_DATA, timestamp=0x0,
+                eventid=0x0, eventslot=False, payload=...)
+        """
+        return """LinFrame(identifier=0x{:x}, echo={}, type={}, timestamp=0x{:x},
+        eventid=0x{:x}, eventslot={}, payload=...)""".format(
```
epage: We should probably not have a multi-line repr

nixnet/types.py
```diff
@@ -220,3 +221,110 @@ def __repr__(self):
...
+    __slots__ = [
+        "identifier",
+        "echo",
+        "type",
+        "timestamp",
+        "eventid",
+        "eventslot",
+        "payload"]
+
+    _FRAME_ID_MASK = 0x000000FF
```
epage: Is this mask correct?

> When Type specifies a LIN frame, this element contains a number in the range 0–63 (inclusive). This number is the LIN frame's ID (unprotected).

http://zone.ni.com/reference/en-XX/help/372841L-01/nixnet/rawframeformat_c/

nixnet/types.py
```diff
@@ -220,3 +221,110 @@ def __repr__(self):
...
+        >>> raw = RawFrame(5, 2, constants.FrameType.LIN_DATA, _cconsts.NX_FRAME_FLAGS_TRANSMIT_ECHO, 0, b'')
+        >>> LinFrame.from_raw(raw)
+        LinFrame(identifier=0x2, echo=True, type=FrameType.LIN_DATA, timestamp=0x5,
+                eventid=0x0, eventslot=False, payload=...)
+        """
+        identifier = frame.identifier & cls._FRAME_ID_MASK
+        lin_frame = LinFrame(identifier, constants.FrameType(frame.type), frame.payload)
+        lin_frame.timestamp = frame.timestamp
+        lin_frame.echo = bool(frame.flags & _cconsts.NX_FRAME_FLAGS_TRANSMIT_ECHO)
+        lin_frame.eventid = frame.info
```
epage: Shouldn't you only read from `frame.info` if `eventslot`  is true?

> For LIN frames, if bit 0 of the Flags element is clear, the Info element is unused (0). If bit 0 of the Flags element is set (event-triggered entry), the Info element contains the event-triggered frame ID, and the Identifier element contains the Unconditional ID from the first payload byte.

nixnet/types.py
```diff
@@ -220,3 +221,110 @@ def __repr__(self):
...
+    """
+
+    __slots__ = [
+        "identifier",
+        "echo",
+        "type",
+        "timestamp",
+        "eventid",
+        "eventslot",
+        "payload"]
```
epage: Reading the documentation, it seems like `eventslot` impacts the meaning of `identifier` and `eventid`

We should at least re-order it so `eventslot` is always referenced before `eventsid` to communicate this.  I'd suggest putting it before `identifier` but that'll mess up `__init__`s defaulting of parameters.

http://zone.ni.com/reference/en-XX/help/372841L-01/nixnet/rawframeformat_c/

nixnet/types.py
```diff
@@ -220,3 +221,110 @@ def __repr__(self):
...
+    _FRAME_ID_MASK = 0x000000FF
+
+    def __init__(self, identifier, type, eventid=b"", payload=b""):
+        # type: (int, constants.FrameType, bytes, bytes) -> None
+        self.identifier = identifier
+        self.echo = False  # Used only for Read
+        self.type = type
+        self.timestamp = 0  # Used only for Read
+        self.eventid = eventid  # Used only for Read
+        self.eventslot = False  # Used only for Read
```
epage: `eventid` is allowed to be passed in but not `eventslot`?  If `eventid` has anything, then `eventslot` should be `True`.

Maybe not include `eventid` in the parameter list and hard code it to `None` so that it works well for people creating a `LinFrame` for writing (where they'll want a payload)

epage:
Note: I made fixes to `__eq__` and `__ne__` in #143.  Please reflect this in your PR.

epage:
```
./nixnet/__init__.py:10:1: F403 'from nixnet.session import *' used; unable to detect undefined names
from nixnet.session import *  # NOQA: F401, F403
^
ERROR: InvocationError: '/home/travis/build/ni/nixnet-python/.tox/flake8/bin/flake8'
```

I submitted a PR that should fix that.  Rebase and re-push.

Also, when you rewrite the history for your PR, we don't get a notification on push so you'll have to let us know when there are changes you want us to look over.

bcornish:
Rebased and repushed as requested.

nixnet/types.py
```diff
@@ -6,6 +6,8 @@
 import collections
 import typing  # NOQA: F401
 
+# flake8: NOQA: E501
+
```
epage: We should isolate suppressions to the lines they impact
1. So we know if they can be removed
2. So we still get reports on other lines

nixnet/_session/frames.py
```diff
@@ -222,7 +222,7 @@ def write(
             frames,
             timeout=10):
         # type: (typing.Iterable[types.Frame], float) -> None
-        """Write raw CAN frame data.
+        """Write raw CAN or LIN frame data.
```
epage: Most of these comments are out of date, sorry for that.

Please change them to be "Read frames" (no reference to "raw", "CAN", or "LIN").

nixnet/types.py
```diff
@@ -388,7 +391,115 @@ def from_raw(cls, frame):
...
+            constants.FrameType.LIN_DATA: LinFrame,
+            constants.FrameType.LIN_BUS_ERROR: LinFrame,
+            constants.FrameType.LIN_NO_RESPONSE: LinFrame,
         }.get(frame.type)
         if frame_type is None:
             raise NotImplementedError("Unsupported frame type", frame.type)
         return frame_type.from_raw(frame)
+
+
+class LinFrame(object):
```
epage: Organizationally, I'd move `LinFrame` before the `XnetFrame`

nixnet/types.py
```diff
@@ -388,7 +391,115 @@ def from_raw(cls, frame):
             constants.FrameType.CANFD_DATA: CanFrame,
             constants.FrameType.CANFDBRS_DATA: CanFrame,
             constants.FrameType.CAN_REMOTE: CanFrame,
+            constants.FrameType.LIN_DATA: LinFrame,
+            constants.FrameType.LIN_BUS_ERROR: LinFrame,
```
epage: Looking at the documentation for `LIN_BUS_ERROR`, it seems like we should create a custom frame for it.
http://zone.ni.com/reference/en-XX/help/372841L-01/nixnet/specialframes_c/

Like CAN, it might be reasonable to defer it.

nixnet/types.py
```diff
@@ -388,7 +391,115 @@ def from_raw(cls, frame):
...
+            constants.FrameType.LIN_NO_RESPONSE: LinFrame,
```
epage: Where is `LIN_NO_RESPONSE` documented?  Trying to decide if it makes sense to map it to a `LinFrame`

bcornish: It's also a special frame: 
[http://zone.ni.com/reference/en-XX/help/372841N-01/nixnet/specialframes/]

Should it be deferred with the rest of the special frame types?

epage: Didn't scroll far enough :)

Let's defer it to a future PR.

nixnet/types.py
bcornish: I moved `LinFrame` before `XnetFrame` and all of the special frames.

epage:
Thinking some more about it, if you add [equality tests](https://github.com/ni/nixnet-python/blob/master/tests/test_frames.py#L128) and increased your [to/from raw coverage](https://coveralls.io/builds/12650406/source?filename=nixnet%2Ftypes.py), this becomes sufficiently tested for what is being added that we might be able to submit it now.

(these tests will need to be added anyways).

nixnet/types.py
```diff
@@ -472,6 +473,111 @@ def __repr__(self):
...
+                self.echo == other_frame.echo,
+                self.type == other_frame.type,
+                self.timestamp == other_frame.timestamp,
+                self.eventslot == other_frame.eventslot,
+                self.eventid == other_frame.eventid,
+                self.payload == other_frame.payload))
+        else:
+            return NotImplemented
+
+    def __repr__(self):
```
epage: FYI in #170, I changed the other frame reprs based on feedback.

bcornish:
Added lin_sched and diagnostic_lin_sched functions. Submitting updates to reprs and equality tests in separate commits.

nixnet/_session/base.py
```diff
@@ -335,6 +335,37 @@ def disconnect_terminals(self, source, destination):
         """
         _funcs.nx_disconnect_terminals(self._handle, source, destination)
 
+    def write_state_lin_schedule_change(self, state_value):
+        # type: () -> None
```
epage: `# type: (typing.Text) -> None`

nixnet/_session/base.py
```diff
@@ -335,6 +335,37 @@ def disconnect_terminals(self, source, destination):
...
+
+        According to the LIN protocol, only the master executes schedules,
+        not slaves. If the
+        :any:`nixnet._session.intf.Interface.lin_master` property is false (slave),
+        this write function implicitly sets that property to true (master). If the
+        interface currently is running as a slave, this write returns an error,
+        because it cannot change to master while running.
+
+        Args:
+            state_value(str): Writes the desired state.
```
epage: You have state_value documented as a `str` but the `nxWriteState` documentation says that it is an int`

http://zone.ni.com/reference/en-XX/help/372841L-01/nixnet/nxwritestate/

nixnet/_session/base.py
```diff
@@ -335,6 +335,37 @@ def disconnect_terminals(self, source, destination):
...
+        """
+        _funcs.nx_write_state(self._handle, constants.WriteState.LIN_SCHEDULE_CHANGE, state_value)
```
epage: `nx_write_state` is expecting `state_value` to be to one of the types defined in `_ctypedefs`.

So 
```python
_funcs.nx_write_state(self._handle, constants.WriteState.LIN_SCHEDULE_CHANGE, _ctypedefs.u32(state_value))
```

nixnet/_session/base.py
```diff
@@ -335,6 +335,37 @@ def disconnect_terminals(self, source, destination):
...
+        According to the LIN protocol, only the master executes schedules,
+        not slaves. If the
+        :any:`nixnet._session.intf.Interface.lin_master` property is false (slave),
+        this write function implicitly sets that property to true (master). If the
+        interface currently is running as a slave, this write returns an error,
+        because it cannot change to master while running.
+
+        Args:
+            state_value(str): Writes the desired state.
+        """
```
epage: > state_value(str): Writes the desired state.

I feel like the parameter name and documentation could be improved.

From the nxWriteState documentation

>  index to the schedule table that the LIN master executes. The schedule tables are sorted the way they are returned from the database with the XNET Cluster Schedules property.

So my recommendation
```
sched_index(int): Index to the schedule table that the LIN master executes.

    The schedule tables are sorted the way they are returned from the database with the XNET Cluster Schedules property.
```

nixnet/_session/base.py
```diff
@@ -335,6 +335,37 @@ def disconnect_terminals(self, source, destination):
...
+        _funcs.nx_write_state(self._handle, constants.WriteState.LIN_SCHEDULE_CHANGE, state_value)
+
+    def write_state_lin_diagnostic_schedule_change(self, state_value):
+        # type: () -> None
```
epage: `# type: (constants.LinDiagnosticSchedule) -> None`

nixnet/_session/base.py
```diff
@@ -335,6 +335,37 @@ def disconnect_terminals(self, source, destination):
...
+        """Writes communication states of an XNET session.
+
+        This function writes a request for the LIN interface to change
+        the diagnostic schedule.
+
+        Args:
+            state_value(str): Writes the desired state.
+        """
+        _funcs.nx_write_state(self._handle, constants.WriteState.LIN_DIAGNOSTIC_SCHEDULE_CHANGE, state_value)
```
epage: Ditto about making it a `u32`

`_ctypedefs.u32(state_value.value)`
(`.value` to convert from enum to int)

nixnet/_session/base.py
```diff
@@ -335,6 +335,37 @@ def disconnect_terminals(self, source, destination):
...
+
+    def write_state_lin_diagnostic_schedule_change(self, state_value):
+        # type: () -> None
+        """Writes communication states of an XNET session.
+
+        This function writes a request for the LIN interface to change
+        the diagnostic schedule.
+
+        Args:
+            state_value(str): Writes the desired state.
```
epage: Again, I think we can do better about the parameter name.  One reason this is important is the parameter names are part of the API definition and changing them breaks compatibility

For example, you can do
```python
session.write_state_lin_diagnostic_schedule_change(state_value=constants.NULL)
```

From reading the documentation, my recommendation
```
schedule(:any:`nixnet._enums.LinDiagnosticSchedule`):  diagnostic schedule that the LIN master executes

File: nixnet/_session/base.py [hunk 22]
```diff
@@ -335,6 +335,37 @@ def disconnect_terminals(self, source, destination):
         """
         _funcs.nx_disconnect_terminals(self._handle, source, destination)
 
+    def write_state_lin_schedule_change(self, state_value):
```
epage: Like the other functions, I don't think we need "write_state" in the name.

So the question is what is a name that makes sense for just this function?

Would `change_lin_schedule` work?

nixnet/_session/base.py
```diff
@@ -335,6 +335,37 @@ def disconnect_terminals(self, source, destination):
...
+        this write function implicitly sets that property to true (master). If the
+        interface currently is running as a slave, this write returns an error,
+        because it cannot change to master while running.
+
+        Args:
+            state_value(str): Writes the desired state.
+        """
+        _funcs.nx_write_state(self._handle, constants.WriteState.LIN_SCHEDULE_CHANGE, state_value)
+
+    def write_state_lin_diagnostic_schedule_change(self, state_value):
```
epage: Like the other functions, I don't think we need "write_state" in the name.

So the question is what is a name that makes sense for just this function?

Would `change_lin_diagnostic_schedule` work?  My main concern is that is a similar name to `change_lin_schedule` but they take completely different types of parameters.

nixnet/_session/base.py
epage: Also, since this will be using a yet-to-be-documented enum, this change should also include documenting the enum.

You can take the documentation from nxWriteState for the values and document the `LinDiagnosticSchedule` class similarly to how we documented the other enums.

nixnet/_session/base.py
bcornish: `change_lin_schedule` makes sense and still conveys the essence of the function (at least from my understanding of it). @lilesj any thoughts?

lilesj: change_lin_schedule or something similar sounds good to me.

epage:
> The addition of LIN examples looks like it's going to require a slight restructuring of the examples.rst so that the CAN and LIN examples are clearly differentiated and can have their own headers. Thoughts?

Was this question from before we talked in person and I suggested us having the CAN/LIN examples in the same files?

tox-integration.ini
```diff
@@ -9,12 +9,14 @@ skip_missing_interpreters=True
 [testenv]
 description = Run tests
 commands =
-    test: pytest --junit-xml=junit/{envname}.xml --junit-prefix={envname} {posargs: --nixnet-in-interface {env:NIXNET_FIXTURE_IN_INTERFACE:CAN1} --nixnet-out-interface {env:NIXNET_FIXTURE_OUT_INTERFACE:CAN2}}
+    test: pytest --junit-xml=junit/{envname}.xml --junit-prefix={envname} {posargs: --can-in-interface {env:CAN_FIXTURE_IN_INTERFACE:CAN1} --can-out-interface {env:CAN_FIXTURE_OUT_INTERFACE:CAN2}}
```
epage: LIN interfaces aren't being passed in on the command line

nixnet/_session/base.py
```diff
@@ -335,6 +335,41 @@ def disconnect_terminals(self, source, destination):
         """
         _funcs.nx_disconnect_terminals(self._handle, source, destination)
 
+    def change_lin_schedule(self, sched_index):
+        # type: (typing.Text) -> None
```
epage: `# type: (int) -> None`

nixnet/_session/intf.py
```diff
@@ -776,6 +776,26 @@ def lin_break_length(self, value):
 
     @property
     def lin_master(self):
+        '''boolean: LIN Master?
```
epage: `# type: () -> bool`

bcornish:
Can you take a look? I believe it's ready to be merged.

PR #144, opened 2017-07-26
feat(API): Simplify read/write calls

epage:
nixnet will now automatically adapt Frame `read`/`write` calls based on
the data being transfered rather than requiring per-frame-type read/write
calls (like `read_can`/`write_can`).

This will help when adding special frame types.  Without this change, they
would have been interpreted as `CanFrame` which would provide weird
results unless someone thought to check `frame.type` and then they'd have
to go through hoops to get it into the correct frame object to interpret
it correctly.

This lowers the visibility of `read_raw`/`write_raw` in the API which
should only be used for some specific advanced cases anyways (arbitrary
data logging).

BREAKING CHANGE:

`read_can` has been replaced by `read`.  If the session reads non-CAN
frames (see special frame types), then those will be included in returned
data but will not be misinterpreted.

`read_raw` has been replaced by `read(..., frame_type=types.RawFrame)`.

`write_can` and `write_frame` have been replaced by `write`.

### DIFF
--- a/nixnet/types.py
+++ b/nixnet/types.py
@@ -2,14 +2,24 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import abc
 import collections
 import typing  # NOQA: F401
 
+import six
+
 from nixnet import _cconsts
 from nixnet import _errors
+from nixnet import _py2
 from nixnet import constants
 
-__all__ = ['DriverVersion', 'CanComm', 'RawFrame', 'CanFrame']
+__all__ = [
+    'DriverVersion',
+    'CanComm',
+    'FrameFactory',
+    'Frame',
+    'RawFrame',
+    'CanFrame']
 
 
 DriverVersion = collections.namedtuple(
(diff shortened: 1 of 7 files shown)


epage:
Since @bcornish might not get back to #142 until tomorrow, I thought I'd go ahead and put this up.

Benefit of earlier: might be able to slip it into 0.1.0 to reduce impact of breakages after it.

Cost of retrofitting #142: abandon the `nixnet/_session/frames.py` changes and just update `XnetFrame` to dispatch the correct `FrameType`s to the new `LinFrame`.

nixnet/types.py
```diff
@@ -230,3 +286,24 @@ def __repr__(self):
...
+        frame_type = {
+            constants.FrameType.CAN_DATA: CanFrame,
+            constants.FrameType.CAN20_DATA: CanFrame,
+            constants.FrameType.CANFD_DATA: CanFrame,
+            constants.FrameType.CANFDBRS_DATA: CanFrame,
+            constants.FrameType.CAN_REMOTE: CanFrame,
+        }.get(frame.type)
+        if frame_type is None:
+            raise NotImplementedError("Unsupported frame type", frame.type)
+        return frame_type.from_raw(frame)
```
epage: @lilesj The purpose of this is for us to create the correct `Frame` object based on the type encoded in the data.  For example, more cases would be added for `LinFrame` and for each of the special frames.

It looked like all of these `FrameType` used the same data format.  Did I get this correct?

lilesj: This looks correct and very scalable.

PR #169, opened 2017-08-02
	 feat(API): Adding partial database API support

### LINKED ISSUE #77: Implement minimal Database API functionality
Implement the following functions from the database API:

- [ ] nxdbAddAlias64 - Allows the user to programmatically create database aliases given a file path
- [ ] nxdbRemoveAlias - Removes a database alias from the system.
- [ ] nxdbGetDatabaseList - Provides a list of aliases on the host PC that can be used by create session. The C function includes the option to return the list of databases on a remote RT system. While that functionality is not currently required the implementation should take this into consideration.
- [ ] nxdbGetDatabaseListSizes - This function returns the buffer size required to hold the entire list of databases on the system. Will probably need to be included in the implementation of nxdbGetDatabaseList.

jashnani:
- [x] Unit test example
- [x] Clean up commits
- [x] Clean up PR
- [x] Integration test
- [x] Include example in docs

Adds partial database API support.

Enhances the API with programmatic database functionality.  This fixed #77.

The example runs as expected. Integration and unit tests run as expected.

### Design Alternatives

Goals
- Be pythonic (including using ABCs that make sense
- Keep things simple to the user; minimizing opening handles unless needed and make it clear when its done

**DatabaseAlias Mapping**
(current design)

`System.databases` is a custom `Mapping` object with Alias objects that store the results.  These objects could also have an "open database" call.

iter, keys, values, items, len, getitem, delitem all work as expected.

setitem could be implemented.  There are a couple of annoyances
- A `DatabaseAlias` object would need to be created by the user and passed in.
- It could only succeed for database formats that have enough information to satisfy NI-XNET, like baud rate.  We'd still need a custom "import" function that accepts additional information that is needed but not included in the database.

**Alias/Path Mapping**

`System.databases` is a custom `Mapping` object from alias to file path.  Everything else is similar to ``DatabaseAlias Mapping``.

**Distinct Functions (no ABC)**

Whether `System` functions or free functions (which would mean they could take the IPAddress for RT support).

### Known Issues
System.databases property is missing a type check because flake8 is throwing an error when the return type is set to 'databases.Databases'.  I suspect this might have to do with 'databases' being overused.

### DIFF
--- a/nixnet/_funcs.py
+++ b/nixnet/_funcs.py
@@ -712,3 +712,45 @@ def nxdb_undeploy(
         database_alias_ctypes,
     )
     _errors.check_for_error(result.value)
+
+
+def nxdb_get_database_list(
+    ip_address,  # type: typing.Text
+    size_of_alias_buffer,  # type: int
+    size_of_filepath_buffer,  # type: int
+):
+    # type: (...) -> typing.Tuple[typing.List[typing.Text], typing.List[typing.Text], int]
+    ip_address_ctypes = _ctypedefs.char_p(ip_address.encode('ascii'))
+    size_of_alias_buffer_ctypes = _ctypedefs.u32(size_of_alias_buffer)
+    size_of_filepath_buffer_ctypes = _ctypedefs.u32(size_of_filepath_buffer)
+    alias_buffer_ctypes = ctypes.create_string_buffer(size_of_alias_buffer)
+    filepath_buffer_ctypes = ctypes.create_string_buffer(size_of_filepath_buffer)
+    number_of_databases_ctypes = _ctypedefs.u32()
+    result = _cfuncs.lib.nxdb_get_database_list(
+        ip_address_ctypes,
+        size_of_alias_buffer_ctypes,
+        alias_buffer_ctypes,
+        size_of_filepath_buffer_ctypes,
+        filepath_buffer_ctypes,
+        number_of_databases_ctypes,
+    )
+    _errors.check_for_error(result.value)
+    alias_buffer = alias_buffer_ctypes.value.decode("ascii")
+    filepath_buffer = filepath_buffer_ctypes.value.decode("ascii")
+    return alias_buffer, filepath_buffer, number_of_databases_ctypes.value
+
+
+def nxdb_get_database_list_sizes(
+    ip_address,  # type: typing.Text
+):
+    # type: (...) -> typing.Tuple[int, int]
+    ip_address_ctypes = _ctypedefs.char_p(ip_address.encode('ascii'))
+    size_of_alias_buffer_ctypes = _ctypedefs.u32()
+    size_of_filepath_buffer_ctypes = _ctypedefs.u32()
+    result = _cfuncs.lib.nxdb_get_database_list_sizes(
+        ip_address_ctypes,
+        size_of_alias_buffer_ctypes,
+        size_of_filepath_buffer_ctypes,
+    )
+    _errors.check_for_error(result.value)
+    return size_of_alias_buffer_ctypes.value, size_of_filepath_buffer_ctypes.value
--- a/nixnet/system/databases.py
+++ b/nixnet/system/databases.py
@@ -1,0 +1,145 @@
+﻿from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import collections
+import typing  # NOQA: F401
+
+import six
+
+from nixnet import _funcs
+
+
+class Databases(collections.Mapping):
+    """Database aliases."""
+
+    def __init__(self, handle):
+        self._handle = handle
+
+    def __repr__(self):
+        return 'System.Databases(handle={0})'.format(self._handle)
+
+    def __get_database_list(self, ip_address):
+        alias_buffer_size, filepath_buffer_size = _funcs.nxdb_get_database_list_sizes(ip_address)
+        aliases, filepaths, _ = _funcs.nxdb_get_database_list(ip_address, alias_buffer_size, filepath_buffer_size)
+        return list(zip(aliases.split(","), filepaths.split(",")))
+
+    def __len__(self):
+        return len(self.__get_database_list(''))
+
+    def __iter__(self):
+        return self.keys()
+
+    def __getitem__(self, index):
+        # type: (str) -> Database
+        """Return the Database object associated with the specified alias.
+
+            Args:
+                index(str): The value of the index (alias name).
+        """
+        if isinstance(index, six.string_types):
+            for alias, filepath in self.__get_database_list(''):
+                if alias == index:
+                    return self._create_item(alias, filepath)
+            else:
+                raise KeyError('Database alias %s not found in the system' % index)
+        else:
+            raise TypeError(index)
+
+    def __delitem__(self, index):
+        # type: (str) -> None
+        """Delete/Remove a database alias from the system.
+
+        This function removes the alias from NI-XNET, but does not affect the
+        database text file. It just removes the alias association to the
+        database file path.
+
+        This function is supported on Windows only, and the alias is removed
+        from Windows only (not RT targets). Use 'undeploy' to remove an alias
+        from a Real-Time (RT) target.
+
+        Args:
+            index(str): The name of the alias to delete.
+        """
+        _funcs.nxdb_remove_alias(index)
+
+    def _create_item(self, database_alias, database_filepath):
+        return Database(database_alias, database_filepath)
+
+    def keys(self):
+        """Return all keys (alias names) used in the Databases object.
+
+            Yields:
+                An iterator to all the keys in the Database object.
+        """
+        for alias, _ in self.__get_database_list(''):
+            yield alias
+
+    def values(self):
+        """Return all Database objects in the system.
+
+            Yields:
+                An iterator to all the values in the Databases object.
+        """
+        for alias, filepath in self.__get_database_list(''):
+            yield self._create_item(alias, filepath)
+
+    def items(self):
+        """Return all aliases and database objects associated with those aliases in the system.
+
+            Yields:
+                An iterator to tuple pairs of alias and database objects in the system.
+        """
+        for alias, filepath in self.__get_database_list(''):
+            yield alias, self._create_item(alias, filepath)
+
+    def add_alias(self, database_alias, database_filepath, default_baud_rate):
+        # type: (str, str, int) -> None
+        """Add a new alias with baud rate size of up to 64 bits to a database file.
+
+        NI-XNET uses alias names for database files. The alias names provide a
+        shorter name for display, allow for changes to the file system without
+        changing the application.
+
+        This function is supported on Windows only.
+
+        Args:
+            database_alias(str): Provides the desired alias name. Unlike the names of
+                other XNET database objects, the alias name can use special
+                characters such as space and dash. Commas are not allowed in the
+                alias name. If the alias name already exists, this function
+                changes the previous filepath to the specified filepath.
+            database_filepath(str): Provides the path to the CANdb, FIBEX, or LDF
+                file. Commas are not allowed in the filepath name.
+            default_baud_rate(int): Provides the default baud rate, used when
+                filepath refers to a CANdb database (.dbc) or an NI-CAN database
+                (.ncd). These database formats are specific to CAN and do not
+                specify a cluster baud rate. Use this default baud rate to
+                specify a default CAN baud rate to use with this alias. If
+                database_filepath refers to a FIBEX database (.xml) or LIN LDF
+                file, the default_baud_rate parameter is ignored. The FIBEX and
+                LDF database formats require a valid baud rate for every
+                cluster, and NI-XNET uses that baud rate as the default.
+        """
+        _funcs.nxdb_add_alias64(database_alias, database_filepath, default_baud_rate)
+
+
+class Database(object):
+    """Database alias."""
+
+    def __init__(
+            self,
+            database_alias,
+            database_filepath,
+    ):
+        self._database_alias = database_alias
+        self._database_filepath = database_filepath
+
+    def __repr__(self):
+        return 'System.Database(alias={}, filepath={})'.format(self._database_alias, self._database_filepath)
+
+    @property
+    def filepath(self):
+        # type: () -> str
+        """str: Get the filepath associated with the Database object"""
+        return self._database_filepath
--- a/nixnet/system/system.py
+++ b/nixnet/system/system.py
@@ -13,6 +13,7 @@ from nixnet import types
 
 from nixnet.system import _device
 from nixnet.system import _interface
+from nixnet.system import databases
 
 
 class System(object):
@@ -21,6 +22,7 @@ class System(object):
         # type: () -> None
         self._handle = None  # To satisfy `__del__` in case nx_system_open throws
         self._handle = _funcs.nx_system_open()
+        self._databases = databases.Databases(self._handle)
 
     def __del__(self):
         if self._handle is not None:
@@ -66,6 +68,11 @@ class System(object):
         _funcs.nx_system_close(self._handle)
 
         self._handle = None
+
+    @property
+    def databases(self):
+        """:any:`nixnet.system.databases.Databases`: Operate on systems's database's aliases"""
+        return self._databases
 
     @property
     def dev_refs(self):
--- a/nixnet_examples/programmatic_database_usage.py
+++ b/nixnet_examples/programmatic_database_usage.py
@@ -1,0 +1,68 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import os
+
+import six
+
+import nixnet
+from nixnet import constants
+from nixnet.system import system
+from nixnet import types
+
+
+def main():
+    with system.System() as my_system:
+        database_alias = 'custom_database'
+        database_filepath = os.path.join(os.path.dirname(__file__), 'nixnet_examples\databases\custom_database.dbc')
+        default_baud_rate = 500000
+        my_system.databases.add_alias(database_alias, database_filepath, default_baud_rate)
+
+    database_name = 'custom_database'
+    cluster_name = 'CAN_Cluster'
+    output_frame = 'CANEventFrame1'
+    interface = 'CAN1'
+
+    with nixnet.FrameOutQueuedSession(
+            interface,
+            database_name,
+            cluster_name,
+            output_frame) as output_session:
+        terminated_cable = six.moves.input('Are you using a terminated cable (Y or N)? ')
+        if terminated_cable.lower() == "y":
+            output_session.intf.can_term = constants.CanTerm.OFF
+        elif terminated_cable.lower() == "n":
+            output_session.intf.can_term = constants.CanTerm.ON
+        else:
+            print("Unrecognised input ({}), assuming 'n'".format(terminated_cable))
+            output_session.intf.can_term = constants.CanTerm.ON
+
+        user_value = six.moves.input('Enter payload [int, int]: ')
+        try:
+            payload_list = [int(x.strip()) for x in user_value.split(",")]
+        except ValueError:
+            payload_list = [2, 4, 8, 16]
+            print('Unrecognized input ({}). Setting data buffer to {}'.format(user_value, payload_list))
+
+        id = types.CanIdentifier(0)
+        payload = bytearray(payload_list)
+        frame = types.CanFrame(id, constants.FrameType.CAN_DATA, payload)
+
+        print("Writing CAN frames using {} alias:".format(database_name))
+
+        i = 0
+        while i < 3:
+            for index, byte in enumerate(payload):
+                payload[index] = byte + i
+
+            frame.payload = payload
+            output_session.frames.write([frame])
+            print('Sent frame with ID %s payload: %s' % (id, payload))
+            i += 1
+
+    with system.System() as my_system:
+        del my_system.databases[database_name]
+
+if __name__ == '__main__':
+    main()
--- a/tests/test_system.py
+++ b/tests/test_system.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import os
 import time
 
 import pytest  # type: ignore
@@ -202,3 +203,29 @@ def test_intf_properties(can_in_interface):
             print(in_intf.dongle_firmware_version)
             print(in_intf.dongle_compatible_revision)
             print(in_intf.dongle_compatible_firmware_version)
+
+
+@pytest.mark.integration
+def test_database_aliases():
+    with system.System() as sys:
+        print(list(sys.databases))
+        print(len(sys.databases))
+        database_alias = 'test_database'
+        dir_name = os.path.dirname(__file__)
+        database_filepath = os.path.join(dir_name, '..', 'nixnet_examples\databases\custom_database.dbc')
+        default_baud_rate = 750000
+        sys.databases.add_alias(database_alias, database_filepath, default_baud_rate)
+        print(len(sys.databases))
+        print(list(sys.databases))
+        print(sys.databases['test_database'].filepath)
+
+        del sys.databases['test_database']
+        print(len(sys.databases))
+        print(list(sys.databases))
+
+        print(list(sys.databases.keys()))
+        print(list(sys.databases.values()))
+        print(list(sys.databases.items()))
+
+        for database in sys.databases:
+            print(database)
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,4 +3,5 @@ include LICENSE
 include nixnet/VERSION
 recursive-include nixnet *.py
 recursive-include nixnet_examples *.py
+include nixnet_examples/databases/custom_database.dbc
 recursive-include tests *.py
--- a/docs/api_reference.rst
+++ b/docs/api_reference.rst
@@ -9,6 +9,7 @@ API Reference
    :caption: Table of Contents:
 
    api_reference/session
+   api_reference/system
    api_reference/convert
    api_reference/constants
    api_reference/types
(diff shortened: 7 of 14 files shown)


jashnani:
Will be posting an update shortly with slight different design (using an alias object).

nixnet_examples/programmatic_database_api.py
```diff
@@ -0,0 +1,91 @@
...
+        print('Adding three database aliases...')
+        database_alias1 = 'custom_database1'
+        database_filepath1 = '../databases/custom_database1.dbc'
+        default_baud_rate1 = '500000'
+        my_system.aliases[database_alias1] = (database_filepath1, default_baud_rate1)
+        
+        database_alias2 = 'custom_database2'
+        database_filepath2 = '../databases/custom_database2.dbc'
+        default_baud_rate2 = '250000'
+        my_system.aliases[database_alias2] = (database_filepath2, default_baud_rate2)
```
epage: I don't think adding through assignment will work for an API like this.  That implies that you'll get the same object back on lookup but we can't implement that.  My recommendation is an explicit add function

nixnet_examples/programmatic_database_api.py
```diff
@@ -0,0 +1,91 @@
...
+        
+        database_alias3 = 'custom_database3'
+        database_filepath3 = '../databases/custom_database3.dbc'
+        default_baud_rate3 = '800000'
+        my_system.aliases[database_alias3] = (database_filepath3, default_baud_rate3)
```
epage: `aliases` doesn't give context to the user of what this is meant to be.  `db_aliases` or `database_aliases` would though.

nixnet_examples/programmatic_database_api.py
```diff
@@ -0,0 +1,91 @@
...
+        i = 0
+        while i < 4:
+            for index, byte in enumerate(payload):
+                payload[index] = byte + i
+
+            frame.payload = payload
+            output_session.frames.write([frame])
+            print('Sent frame with ID %s payload: %s' % (id, payload))
+
+def main():
```
epage: I'm assuming this is more so to showcase the API work you are doing rather than an example because I wouldn't expect an example to be creating multiple aliases.

nixnet_examples/programmatic_database_api.py
```diff
@@ -0,0 +1,91 @@
...
+            database_name,
+            cluster_name,
+            output_frame) as output_session:
+        terminated_cable = six.moves.input('Are you using a terminated cable (Y or N)? ')
+        if terminated_cable.lower() == "y":
+            output_session.intf.can_term = constants.CanTerm.OFF
+        elif terminated_cable.lower() == "n":
+            output_session.intf.can_term = constants.CanTerm.ON
+        else:
+            print("Unrecognised input ({}), assuming 'n'".format(terminated_cable))
```
epage: If you're making this a function distinct function then it shouldn't take user input but should receive all parameters from the caller.

nixnet_examples/programmatic_database_api.py
```diff
@@ -0,0 +1,91 @@
+from __future__ import absolute_import
```
epage: `programmatic_database_api.py`

The name feels very awkward to me.  Is there an equivalent C or LV example for what this is trying to accomplish?  What is it named?  What does it look like / do?

lilesj: Agreed, the functionality is limited to adding and removing aliases for the moment so the name should probably reflect that in some way.

jashnani: I couldn't find an equivalent example for what I'm trying to do. There is a C example called XNET_Database_Browser but it doesn't leverage add or remove alias. It might be good for the get database list or get database list size functions that I'm writing.

I trying to demonstrate that the same read/write code can be reused with different database aliases (different cluster, baud rates, etc.). This sounded like a common use case for aliases based on my conversation with Jeff.

How about `dynamic_database_usage.py` or `programatic_database_usage.py`?

nixnet_examples/programmatic_database_api.py
jashnani: No, I was expecting us to create multiple aliases and use them in the example. @lilesj what are your thoughts on this? are there other uses cases/workflows I should consider writing examples for?

nixnet_examples/programmatic_database_api.py
```diff
@@ -0,0 +1,95 @@
...
+        alias2 = Alias(database_alias2, database_filepath2, default_baud_rate2)
+        my_system.aliases[alias2.database_name] = alias2
+        
+        database_alias3 = 'custom_database3'
+        database_filepath3 = '../databases/custom_database3.dbc'
+        default_baud_rate3 = '800000'
+        alias3 = Alias(database_alias3, database_filepath3, default_baud_rate3)
+        my_system.aliases[alias3.database_name] = alias3
+        print(list(my_system.aliases))
+
```
lilesj: Since the intent of the example is to demonstrate the use of the database functions, can we call the alias functions first to show they are being created in preparation for the session objects.

nixnet_examples/programmatic_database_api.py
lilesj: Demonstrating how to parse a folder for DB files is an important use case. We might want to change one of the database file to a fibex file to show we can do it with multiple file types. My comment below mentions how we can reinforce the purpose of the example by reordering the operations a bit.

nixnet_examples/programmatic_database_api.py
```diff
@@ -0,0 +1,74 @@
...
+pp = pprint.PrettyPrinter(indent=4)
+
+
+def main():
+    with system.System() as my_system:
+        print(list(my_system.databases))
+        database_alias = 'custom_database'
+        database_filepath = '../databases/custom_database.dbc'
+        default_baud_rate = '500000'
+        my_system.databases.add(database_alias, database_filepath, default_baud_rate)
```
epage: I thought we were leaning towards this being called `add_alias`?

nixnet_examples/programmatic_database_api.py
epage: Resolved

nixnet_examples/programmatic_database_api.py
```diff
@@ -0,0 +1,74 @@
...
+from nixnet import types
+from nixnet.system import system
+
+
+pp = pprint.PrettyPrinter(indent=4)
+
+
+def main():
+    with system.System() as my_system:
+        print(list(my_system.databases))
```
epage: Hmm, I'm torn.

Technically, when you iterate on a mapping, you get back the keys rather than the values.

DAQmx implemented a [sequence](https://github.com/ni/nidaqmx-python/blob/master/nidaqmx/_task_modules/channel_collection.py) that also accepts strings.  Not sure how well that fits here.

I'd say carry on but this is something we need to eventually resolve.

epage: Opened #173 about this.

The interesting aspects for this situation
- Can we support integer indexes?  This requires the database list returned by the driver to be a stable order
- Is it ok if we still implement get/del?  Since get is just a function, probably.  del is fine; it works even on mutable sequences (granted, we won't have a fully mutable sequence).

nixnet_examples/programmatic_database_api.py
jashnani: I changed it in the implementation but forgot to update the example.

jashnani:
Shortly, I will post a separate commit with unit/system tests that should go along with this feature.

nixnet/system/databases.py
```diff
@@ -0,0 +1,147 @@
+﻿from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import collections
+import itertools
+import six
+import typing  # NOQA: F401
```
epage: `six` is a third party module, please have it in a separate section

nixnet/_funcs.py
```diff
@@ -695,3 +695,43 @@ def nxdb_undeploy(
...
+        size_of_alias_buffer_ctypes,
+        alias_buffer_ctypes,
+        size_of_filepath_buffer_ctypes,
+        filepath_buffer_ctypes,
+        number_of_databases_ctypes,
+    )
+    _errors.check_for_error(result.value)
+    alias_list = alias_buffer_ctypes.value.decode("ascii").split(",")
+    filepath_list = filepath_buffer_ctypes.value.decode("ascii").split(",")
+    return list(zip(alias_list, filepath_list))
```
epage: We've generally left post-processing like this to the clients of `_funcs`

nixnet/_funcs.py
```diff
@@ -695,3 +695,43 @@ def nxdb_undeploy(
...
     )
     _errors.check_for_error(result.value)
+
+
+def nxdb_get_database_list(
+    ip_address,  # type: typing.Text
+):
+    # type: (...) -> typing.Tuple[List[typing.Text], List[typing.Text], int]
+    ip_address_ctypes = _ctypedefs.char_p(ip_address.encode('ascii'))
+    size_of_alias_buffer, size_of_filepath_buffer = nxdb_get_database_list_sizes(ip_address)
```
epage: Generally we leave policy like this to the callers of `_funcs`.  If, in the future, we add code-gen, it'll make life a lot easier

nixnet/system/databases.py
```diff
@@ -0,0 +1,147 @@
...
+        self._handle = handle
+
+    def __repr__(self):
+        return 'System.Databases(handle={0})'.format(self._handle)
+
+    def __len__(self):
+        return len(_funcs.nxdb_get_database_list(''))
+
+    def __iter__(self):
+        for alias, filepath in _funcs.nxdb_get_database_list(''):
```
epage: `for alias, _ in ...`

`_` is a generally recommended placeholder for "don't care" values

nixnet/system/databases.py
```diff
@@ -0,0 +1,147 @@
...
+                index(str): The value of the index (alias name).
+        """
+        if isinstance(index, six.string_types):
+            for alias, filepath in _funcs.nxdb_get_database_list(''):
+                if alias == index:
+                    return self._create_item(self._handle, index, alias, filepath)
+            else:
+                raise ValueError('Database alias %s not found in the system' % index)
+        else:
+            raise KeyError(index)
```
epage: I might not have been clear before.

We should throw `KeyError` when the key isn't found.

As for what to throw on unsupported types, it looks like we've set the precedence of `TypeError`
https://github.com/ni/nixnet-python/blob/master/nixnet/_session/collection.py#L44

nixnet/system/databases.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def __repr__(self):
+        return 'System.Databases(handle={0})'.format(self._handle)
+
+    def __len__(self):
+        return len(_funcs.nxdb_get_database_list(''))
+
+    def __iter__(self):
+        for alias, filepath in _funcs.nxdb_get_database_list(''):
+            yield alias
```
epage: Also, why not have `__iter__` implemented by calling `keys`?  It'll make the intent more clear

nixnet/system/databases.py
```diff
@@ -0,0 +1,147 @@
...
+        alias name in any application for the Windows host and RT target.
+
+        Args:
+            database_alias(str): Provides the desired alias name. Unlike the names of
+                other XNET database objects, the alias name can use special
+                characters such as space and dash. Commas are not allowed in the
+                alias name. If the alias name already exists, this function
+                changes the previous filepath to the specified filepath.
+            database_filepath(str): Provides the path to the CANdb, FIBEX, or LDF
+                file. Commas are not allowed in the alias name.
```
epage: This comment is bad.  Its talking about aliases

nixnet/system/databases.py
```diff
@@ -0,0 +1,147 @@
...
+            default_baud_rate(int):
+                Provides the default baud rate, used when filepath refers to a
+                CANdb database (.dbc) or an NI-CAN database (.ncd). These
+                database formats are specific to CAN and do not specify a cluster
+                baud rate. Use this default baud rate to specify a default CAN
+                baud rate to use with this alias. If database_filepath refers to
+                a FIBEX database (.xml) or LIN LDF file, the default_baud_rate
+                parameter is ignored. The FIBEX and LDF database formats require
+                a valid baud rate for every cluster, and NI-XNET uses that baud
+                rate as the default.
```
epage: I think we talked about defaulting `default_baud_rate` to `None`.

- If its `None` and a file type that doesn't need it, we just pass 0.
- If its `None` and is file type that needs it, we throw.
- If its not `None` and a file type that doesn't need it, we error.

Granted, there are pros/cons to this.  I'm still torn on it myself.

nixnet/system/databases.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def __init__(
+            self,
+            handle,
+            index,
+            database_alias,
+            database_filepath,
+    ):
+        self._handle = handle
+        self._index = index
```
epage: Why does this take a handle or index?

nixnet/system/databases.py
```diff
@@ -0,0 +1,147 @@
...
+        self._database_alias = database_alias
+        self._database_filepath = database_filepath
```
epage: Don't we need to define a property to expose `filepath`?

nixnet_examples/programmatic_database_api.py
epage: This hasn't been renamed yet

nixnet_examples/programmatic_database_api.py
```diff
@@ -0,0 +1,81 @@
...
+        print(list(my_system.databases.keys()))
+        print(list(my_system.databases.values()))
+        print(list(my_system.databases.items()))
+        print(len(my_system.databases))
```
epage: Didn't we discuss removing all these prints?  This is an example and not a test

nixnet_examples/programmatic_database_api.py
```diff
@@ -0,0 +1,81 @@
...
+        database_alias = 'custom_database'
+        database_filepath = 'databases/custom_database.dbc'
```
epage: Probably want to prefix this with `basename(__file__)` or whatever the call would be so they can run this from other locations

nixnet/system/databases.py
jashnani: Can you elaborate on what's bad? I can't quite tell by reading the Args docs.

epage: Its not a syntax thing

>  database_filepath(str): ... Commas are not allowed in the alias name.

nixnet/system/databases.py
jashnani: Fixed.

nixnet/system/databases.py
jashnani: Fixed.

nixnet/system/databases.py
jashnani: Fixed.

nixnet/system/databases.py
jashnani: Fixed.

nixnet/system/databases.py
jashnani: Fixed.

nixnet/system/databases.py
jashnani: Yes, I've a added a property for this. Just the get function for now.

nixnet/_funcs.py
jashnani: I've cleaned this up and move the call the sizes function one layer above in databases.py

nixnet/_funcs.py
jashnani: I left the decoding of the strings logic in this layer thinking that this would need to be generated for all string output types anyway.

I've moved  the 'split' and 'zip' logic to the layer above (databases) though.

nixnet_examples/programmatic_database_api.py
jashnani: Yes, I just forgot to take them out. Good catch.

nixnet/system/databases.py
jashnani: Got rid of them, don't need them anymore.

nixnet/system/databases.py
jashnani: It feels weird to error on an optional argument. I'm just leaving it as is and let the driver ignore/throw accordingly.

nixnet/system/databases.py
```diff
@@ -0,0 +1,149 @@
...
+    def __init__(self, handle):
+        self._handle = handle
+
+    def __repr__(self):
+        return 'System.Databases(handle={0})'.format(self._handle)
+
+    def __get_database_list(self, ip_address):
+        alias_buffer_size, filepath_buffer_size = _funcs.nxdb_get_database_list_sizes(ip_address)
+        aliases, filepaths, _ = _funcs.nxdb_get_database_list(ip_address, alias_buffer_size, filepath_buffer_size)
+        return list(zip(aliases.split(","), filepaths.split(",")))
```
epage: Should we be returning `number_of_databases` to `__len__` instead of doing a string split, zip, and list?

nixnet/system/system.py
```diff
@@ -68,6 +70,11 @@ def close(self):
         self._handle = None
 
     @property
+    def databases(self):
```
epage: This doesn't have a type annotation

nixnet/system/databases.py
```diff
@@ -0,0 +1,149 @@
...
+        # type: (str, str, int) -> None
+        """Add a new alias with baud rate size of up to 64 bits to a database file.
+
+        NI-XNET uses alias names for database files. The alias names provide a
+        shorter name for display, allow for changes to the file system without
+        changing the application, and enable efficient deployment to LabVIEW
+        Real-Time (RT) targets.
+
+        This function is supported on Windows only. For RT targets, you can pass
+        the new alias to 'deploy' to transfer an optimized binary image of the
```
epage: Kill all the RT stuff since we're not really supporting those use cases with this API (atm)

nixnet_examples/programmatic_database_api.py
epage: resolvved

nixnet/system/databases.py
epage: resolved.

nixnet/system/databases.py
epage: resolved.

nixnet/system/databases.py
epage: Resolved.

nixnet/system/databases.py
epage: Resolved.

nixnet/system/databases.py
epage: Resolved.

nixnet/system/databases.py
epage: Its not quite weird to have conditionally optional parameters.

nixnet/system/databases.py
epage: Resolved.

nixnet/system/databases.py
epage: Resolved.

nixnet/_funcs.py
epage: Resolved.

nixnet/_funcs.py
epage: Resolved.

nixnet_examples/programmatic_database_api.py
epage: Resolved.

tests/test_system.py
```diff
@@ -202,3 +202,24 @@ def test_intf_properties(can_in_interface):
...
             print(in_intf.dongle_compatible_firmware_version)
+
+
+@pytest.mark.integration
+def test_database_aliases():
+    with system.System() as sys:
+        print(list(sys.databases))
+        num_databases = len(sys.databases)
+        database_alias = 'test_database'
+        database_filepath = 'nixnet_examples/databases/custom_database.dbc'
```
epage: Are we sure that the CWD will always be correct for this?

nixnet_examples/programmatic_database_api.py
epage: This doesn't look like its resolved.

nixnet/system/databases.py
```diff
@@ -0,0 +1,149 @@
...
+
+import collections
+import typing  # NOQA: F401
+
+import six
+
+from nixnet import _funcs
+
+
+class Databases(collections.Mapping):
```
epage: Please remove relevant functions from `nxdb.py`

https://github.com/ni/nixnet-python/blob/master/nixnet/nxdb.py

nixnet_examples/programmatic_database_usage.py
```diff
@@ -0,0 +1,70 @@
...
+            database_name,
+            cluster_name,
+            output_frame) as output_session:
+        terminated_cable = six.moves.input('Are you using a terminated cable (Y or N)? ')
+        if terminated_cable.lower() == "y":
+            output_session.intf.can_term = constants.CanTerm.OFF
+        elif terminated_cable.lower() == "n":
+            output_session.intf.can_term = constants.CanTerm.ON
+        else:
+            print("Unrecognised input ({}), assuming 'n'".format(terminated_cable))
```
epage: You don't have coverage on this case

nixnet/system/databases.py
```diff
@@ -0,0 +1,149 @@
...
+            database_filepath,
+    ):
+        self._database_alias = database_alias
+        self._database_filepath = database_filepath
+
+    def __repr__(self):
+        return 'System.Database(alias={}, filepath={})'.format(self._database_alias, self._database_filepath)
+
+    @property
+    def filepath(self):
```
epage: This property is not tested

nixnet/system/databases.py
```diff
@@ -0,0 +1,149 @@
...
+        aliases, filepaths, _ = _funcs.nxdb_get_database_list(ip_address, alias_buffer_size, filepath_buffer_size)
+        return list(zip(aliases.split(","), filepaths.split(",")))
+
+    def __len__(self):
+        return len(self.__get_database_list(''))
+
+    def __iter__(self):
+        return self.keys()
+
+    def __getitem__(self, index):
```
epage: This function is not tested

nixnet/system/databases.py
```diff
@@ -0,0 +1,149 @@
...
+from nixnet import _funcs
+
+
+class Databases(collections.Mapping):
+    """Database aliases."""
+
+    def __init__(self, handle):
+        self._handle = handle
+
+    def __repr__(self):
```
epage: This function is not tested

nixnet_examples/programmatic_database_usage.py
jashnani: Github code selection/snippet is terrible. Can you clarify what part?

tests/test_system.py
jashnani: What is CWD?

nixnet/system/system.py
jashnani: I was getting a flake8 error when trying to have return type: databases.Databases

So I removed it. It may have to do with 'databases' being overloaded

tests/test_system.py
epage: Current Working Directory

nixnet/system/system.py
epage: Possibly but we should dig deeper or call these thing out in the PR rather than deleting.

nixnet_examples/programmatic_database_usage.py
epage: The `else` branch.  You can also look it up yourself in the coverage reports.

nixnet/system/databases.py
jashnani: I don't think it makes a difference.

nixnet/system/databases.py
jashnani: How are we tracking what documentation needs to be added where when RT is supported?

tests/test_system.py
jashnani: I'm not sure.

nixnet/system/databases.py
epage: I generally expected "len" operations to be "fast" but in this case we're doing more post-processing than is needed.

tests/test_system.py
epage: So should we make the path an absolute path?

nixnet_examples/programmatic_database_api.py
```diff
@@ -64,26 +65,29 @@ def main():
         database_alias1 = 'custom_database1'
         database_filepath1 = '../databases/custom_database1.dbc'
         default_baud_rate1 = '500000'
-        my_system.aliases[database_alias1] = (database_filepath1, default_baud_rate1)
+        alias1 = Alias(database_alias1, database_filepath1, default_baud_rate1)
+        my_system.aliases[alias1.database_name] = alias1
```
epage: resolved.

nixnet_examples/programmatic_database_api.py
epage: Oops, I posted another comment on this.

nixnet/system/databases.py
jashnani: I don't think doing a split, zip and list is going to be that expensive. Especially when we don't expect users to use a ton of aliases at once. Jeff expects around 100 as a maximum.

Leaving it as is.

nixnet/system/system.py
jashnani: It's just a type annotation! I don't think it's super critical that needs to be addressed right away, adding notes to the PR

nixnet/system/databases.py
epage: Resolved

tests/test_system.py
```diff
@@ -202,3 +203,28 @@ def test_intf_properties(can_in_interface):
...
             print(in_intf.dongle_compatible_firmware_version)
+
+
+@pytest.mark.integration
+def test_database_aliases():
+    with system.System() as sys:
+        print(list(sys.databases))
+        num_databases = len(sys.databases)
+        database_alias = 'test_database'
+        database_filepath = os.path.join(os.path.dirname(__file__), '..', 'nixnet_examples\databases\custom_database.dbc')
```
epage: For clean up after this, we should have another `dirname` rather than `..`

Issue #170, opened 2017-08-03
Printing a frame object does not print payload values


KilometersG:
I reproduced this issue using the can_frame_queued_io.py example. My only modifications were to remove the __future__ and six imports and use the built in functions. 

When I print a frame object I get the full information on the frame including Arb ID, Echo setting, Type, and timestamp, but the printed string always ends with "payload=...)".

I have reproduced this with both PrettyPrinter and the default print function. I confirmed that if you print frame.payload the full payload information does print. I also tried printing a string longer then the CAN frame to make sure my IDE didn't have a setting to cut off strings beyond a certain length (it does not cut off long strings).

- OS: Windows 7 Professional 64-bit
- Python version: 3.5.3 
- IDE: Pycharm Community Edition 2017.2
- NI-XNET version: 17.0
- Hardware: PXIe-8510/6 

*TODO:* paste the output from
```bash
$ python -m pip list
```
enum34 <1.1.6>
nixnet <0.1.0>
pip <9.0.1>
setuptools <28.8.0>
six <1.10.0>
typing <3.6.1>

epage:
Thank you for providing feedback on this.

I was concerned about how big the payload might be (like with CAN FD or J1939) and decided to leave it off (see [types.py](https://github.com/ni/nixnet-python/blob/master/nixnet/types.py#L360)).  I assumed it'd be better for people to print the payload separately rather than getting overwhelmingly large output text that in aggregate might obscure the more critical information.

Some ideas I've come up with for this
- Have `__repr__` only print echo/type if it is non-default
  - Includes defining a default for `type` (`CAN_DATA` is probably best)
  - Unfortunately, printing lines of frames won't align as well
- Include the payload but cap its size, ending with a `...` if its too big
  - Maybe if it goes above 8 bytes, show 5 + `...` (so it still totals 8).

Any thoughts on the trade offs of those?  Any other ideas (including just ignoring the length of the `repr` and just showing it all)?

KilometersG:
I'm trying to think through customer expectations. The reason I posted this as an issue is because this is in an example that's setup to do a loopback test. I would expect a loopback test to show the payload sent and payload received to verify proper communication.  In the short term, simply adding a print(frame.payload) to the examples would work.

Long term I like your second suggestion. If that was the case I would say it's important that we document in an easy to find place something along the lines of "For full payload readout print frame.payload" or something like that. 

[...]

PR #174, opened 2017-08-04
fix(examples): Improve reporting of frames (+ other stuff)

epage:
This mostly focuses on fixing #170. 
- Improve the frame repr
- Print payloads in examples

Some other minor fixes are done while at it.

### DIFF
--- a/nixnet/types.py
+++ b/nixnet/types.py
@@ -271,14 +271,24 @@ class RawFrame(Frame):
         """RawFrame debug representation.
 
         >>> RawFrame(1, 2, constants.FrameType.CAN_DATA, 3, 4)
-        RawFrame(timestamp=0x1, identifier=0x2, type=FrameType.CAN_DATA, flags=0x3, info=0x4, payload=...)
+        RawFrame(timestamp=0x1, identifier=0x2, type=FrameType.CAN_DATA, flags=0x3, info=0x4)
         """
-        return "RawFrame(timestamp=0x{:x}, identifier=0x{:x}, type={}, flags=0x{:x}, info=0x{:x}, payload=...)".format(
+        optional = []
+        if self.flags != 0:
+            optional.append('flags=0x{:x}'.format(self.flags))
+        if self.info != 0:
+            optional.append('info=0x{:x}'.format(self.info))
+        if self.payload:
+            optional.append('len(payload)={}'.format(len(self.payload)))
+        if optional:
+            optional_params = ', {}'.format(", ".join(optional))
+        else:
+            optional_params = ''
+        return "RawFrame(timestamp=0x{:x}, identifier=0x{:x}, type={}{})".format(
             self.timestamp,
             self.identifier,
             self.type,
-            self.flags,
-            self.info)
+            optional_params)
 
 
 class CanFrame(Frame):
@@ -317,7 +327,7 @@ class CanFrame(Frame):
 
         >>> raw = RawFrame(5, 0x20000001, constants.FrameType.CAN_DATA, _cconsts.NX_FRAME_FLAGS_TRANSMIT_ECHO, 0, b'')
         >>> CanFrame.from_raw(raw)
-        CanFrame(CanIdentifier(0x1, extended=True), echo=True, type=FrameType.CAN_DATA, timestamp=0x5, payload=...)
+        CanFrame(CanIdentifier(0x1, extended=True), echo=True, timestamp=0x5)
         """
         identifier = CanIdentifier.from_raw(frame.identifier)
         can_frame = CanFrame(identifier, constants.FrameType(frame.type), frame.payload)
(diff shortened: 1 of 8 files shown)


nixnet/types.py
```diff
@@ -231,7 +231,7 @@ class RawFrame(Frame):
         "info",
         "payload"]
 
-    def __init__(self, timestamp, identifier, type, flags, info, payload=b""):
+    def __init__(self, timestamp, identifier, type, flags=0, info=0, payload=b""):
```
epage: @KilometersG and @lilesj, do these defaults make sense?

nixnet/types.py
```diff
@@ -300,7 +310,7 @@ class CanFrame(Frame):
         "timestamp",
         "payload"]
 
-    def __init__(self, identifier, type, payload=b""):
+    def __init__(self, identifier, type=constants.FrameType.CAN_DATA, payload=b""):
```
epage: @KilometersG and @lilesj, do these defaults make sense?

nixnet/types.py
```diff
@@ -271,14 +271,24 @@ def __repr__(self):
...
+            optional_params = ', {}'.format(", ".join(optional))
+        else:
+            optional_params = ''
+        return "RawFrame(timestamp=0x{:x}, identifier=0x{:x}, type={}{})".format(
             self.timestamp,
             self.identifier,
             self.type,
-            self.flags,
-            self.info)
+            optional_params)
```
epage: @KilometersG and @AndrewHeim, thoughts on the new repr?

nixnet/types.py
```diff
@@ -361,14 +371,27 @@ def __repr__(self):
...
+        if optional:
+            optional_params = ', {}'.format(", ".join(optional))
+        else:
+            optional_params = ''
+        return "CanFrame({}{})".format(
             self.identifier,
-            self.echo,
-            self.type,
-            self.timestamp)
+            optional_params)
```
epage: @KilometersG and @AndrewHeim, thoughts on the new repr?

nixnet_examples/can_frame_queued_io.py
```diff
@@ -75,8 +72,8 @@ def main():
                 count = 1
                 frames = input_session.frames.read(count)
                 for frame in frames:
-                    print('Received frame: ')
-                    pp.pprint(frame)
+                    print('Received frame: {}'.format(frame))
+                    print('    payload={}'.format(list(six.iterbytes(frame.payload))))
```
epage: @KilometersG and @AndrewHeim, thoughts on the new example output?

nixnet/types.py
AndrewHeim: I like it.

I'm not certain why we need the optional flags.  My initial thought was to print all of them.  But, then I read the [definition of repr: ](https://docs.python.org/2/library/repr.html)
class repr.Repr

    ...size limits for different object types are added to avoid the generation of representations which are excessively long.

You're right about length.  So, I'm okay with the optional flags if you guys want them.

nixnet/types.py
AndrewHeim: Like

nixnet_examples/can_frame_queued_io.py
AndrewHeim: Like

nixnet/types.py
epage: I've been going back and forth on the optional.

As long as payload only shows the length, we might be able to get away without the optional parameters.  

The value of hiding the default fields is it makes meaningful information standout more.  My only concern is alignment between rows of frames being off, making it harder to spot specific differences (a missing field will be obvious of course).  I'm assuming that when comparing frames, they will most likely be similar to each other wrt which fields are optional, so they'll probably still line up.

epage:
```python3
>>> CanBusErrorFrame(100, constants.CanCommState.BUS_OFF, True, constants.CanLastErr.STUFF, 1, 2).to_raw()
Expected:
    RawFrame(timestamp=0x64, identifier=0x0, type=FrameType.CAN_BUS_ERROR, len(payload)=5)
Got:
    RawFrame(timestamp=0x64, identifier=0x0, type=FrameType.CAN_BUS_ERROR, len(payload)=15)
```

This failed in Travis on Python 2.7.13 and Python 2.7.12 (aff251e54385, Nov 09 2016, 18:02:49) [PyPy 5.6.0 with GCC 4.8.2].

I cannot reproduce on my system with Python 2.7.11.

I looked at other areas of the code where we call `bytes()` but it is mostly there to convert a `bytearray` to `bytes` if needed.

*sigh* Python 2.7 has way too many differences in behavior from update to update.

nixnet/types.py
epage: Talked to Jeff and he was fine with this

nixnet/types.py
epage: Talked to Jeff and he was fine with this

PR #176, opened 2017-08-04
feat(API): Support read/write signals X-Y

epage:
TODO
- [ ] Make example work
- [ ] Integration test
- [ ] Decide if its `XY` or `Xy` in PascalCase.


nixnet/session.py
```diff
@@ -21,7 +21,9 @@
     "FrameInSinglePointSession",
     "FrameOutSinglePointSession",
     "SignalInSinglePointSession",
-    "SignalOutSinglePointSession"]
+    "SignalOutSinglePointSession",
+    "SignalInXYSession",
+    "SignalOutXYSession"]
```
epage: Should it be `SignalInXYSession` or `SignalInXySession`?

PEP-8 only says to use [CapWords](https://www.python.org/dev/peps/pep-0008/#class-names).  In newer python stdlib code, [ABCMeta](https://docs.python.org/3/library/abc.html#abc.ABCMeta) caps each part of an acronym.

So far we've been following the convention of not putting acronyms in caps. 

Does `XY` count as an acronym (`Xy`) or two separate "words" (`XY`)?

bcornish: So I read "XY" as sort of "X and Y" which would make me inclined to lean more towards `XY`.

salehcDA:
Are there any plans to continue development on this?

PR #223, opened 2018-01-31
 refactor(API): Rename 'db' namespace to 'database'

### LINKED ISSUE #221: Consider renaming the nixnet.db namespace to nixnet.database
I'm increasingly tempted to suggest we do this rename based on the following:

1. When I look at existing namespaces and class names (http://nixnet.readthedocs.io/en/latest/api_reference.html), I only see one abbreviated.
2. When I compare things with the C API (nixnet.h), almost every abbreviation to 'db' is some variation of the prefix nx*db. When not used as a function or error prefix, 'database' is almost always spelled out explicitly.

As I understand it, renaming the namespace involves renaming the 'db' folder to 'database', which involves two options:

1. Renaming the folder and fixing code in a single push, which probably causes some git history (blame) to be lost.
2. Renaming the folder first, pushing, then pushing again the fixed namespaces. This may keep history but break git-bisect (binary search for bugs) in the case where it happens to land on the broken push.

Any thoughts on (a) whether to do the rename and (b) which rename method to proceed with?

FYI, I feel keeping history would be more valuable than worrying about an unlikely-to-hit breakage in git-bisect.

d-bohls:
Fixes #221

BREAKING CHANGE:

Rename 'db' namespace to 'database'

- [X] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nixnet-python/blob/master/CONTRIBUTING.rst).
- [X] New tests have been created for any new features or regression tests for bugfixes.
- [X] `tox` successfully runs, including unit tests and style checks (see [CONTRIBUTING.md](https://github.com/ni/nixnet-python/blob/master/CONTRIBUTING.rst)).


nixnet/_session/intf.py
```diff
@@ -8,7 +8,7 @@
 
 from nixnet import _props
 from nixnet import constants
-from nixnet.db import _frame
+from nixnet.database import _frame
```
epage: I was surprised at first that we didn't need to update the rst file for this but we haven't created it yet

https://github.com/ni/nixnet-python/tree/master/docs/api_reference

d-bohls: I actually started creating the rst and that's exactly what prompted me to want this done before I start churning out database documentation.

nixnet/database/_frame.py
```diff
@@ -21,7 +21,7 @@ def __init__(self, handle):
         self._mux_static_signals = _collection.DbCollection(
             self._handle, constants.ObjectClass.SIGNAL, _cconsts.NX_PROP_FRM_MUX_STATIC_SIG_REFS, _signal.Signal)
         self._mux_subframes = _collection.DbCollection(
-            self._handle, constants.ObjectClass.SUBFRAME, _cconsts.NX_PROP_PDU_MUX_SUBFRAME_REFS, _subframe.SubFrame)
+            self._handle, constants.ObjectClass.SUBFRAME, _cconsts.NX_PROP_FRM_MUX_SUBFRAME_REFS, _subframe.SubFrame)
```
jashnani: Was this a bug?

d-bohls: Yes.

PR #225, opened 2018-02-06
chore(examples): Rename in-memory db in examples.

d-bohls:
In our examples, make it clear that in-memory
databases are not restricted to ':memory:'.

Also, it's probably better to avoid using
':memory:' for in-memory databases in case
that might cause conflict with the stream sessions.

### DIFF
--- a/nixnet_examples/can_dynamic_database_creation.py
+++ b/nixnet_examples/can_dynamic_database_creation.py
@@ -12,7 +12,7 @@ from nixnet import database
 
 
 def main():
-    database_name = ':memory:'
+    database_name = ':CAN_Database:'
     cluster_name = 'CAN_Cluster'
     frame_name = 'CAN_Event_Frame'
     signal_1_name = 'CAN_Event_Signal_1'
@@ -21,7 +21,7 @@ def main():
     output_interface = 'CAN1'
     input_interface = 'CAN2'
 
-    # Open the default in-memory database.
+    # Open an in-memory database.
     with database.Database(database_name) as db:
 
         # Add a CAN cluster, a frame, and two signals to the database.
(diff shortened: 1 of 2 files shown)


nixnet_examples/can_dynamic_database_creation.py
```diff
@@ -12,7 +12,7 @@
 
 
 def main():
-    database_name = ':memory:'
+    database_name = ':CAN_Database:'
```
jashnani: I'd prefer we use `:memory:` in our examples unless we have two or more in-memory databases being used at the same time. I don't think we should promote using random strings between the colon because there is a convention for this. See [database documentation](http://zone.ni.com/reference/en-XX/help/372841L-01/nixnet/databaseprogramming/), section Multiple Databases Simultaneously.

"NI-XNET allows opening up to seven distinct databases at the same time. You can open any database from a database file or in memory. To open multiple in-memory databases, use the name :memory[<digit>]:; for example, :memory:, :memory1:, :memory2:."

Maybe we can create a new example showing that you can edit/use multiple in-memory databases at the same time. Thoughts?

d-bohls: I see your point that the documentation/examples seem to encourage the user to use ":memory[n]:" naming for in-memory databases. I'm left wondering if this is really a best-practice or just something like an artifact of brevity in the documentation. I'd like to look into that further.

From my discussion with Jason Smith, I was left with the impression a decisively named database would be desirable to potentially avoid accidental collisions with another thread that may happen to be using the same database name. For all I know, this situation may be unlikely given how customers use XNET, but it seems possible from what Jason said. I'll follow up on this.

sjasonsmith: It probably is best to stick with the naming that is mentioned in the NI-XNET documentation.
The most common use for in-memory databases is going to be stream sessions where the database contents aren't extremely critical. For more advanced uses, I would expect most people to be opening databases from disk.

PR #227, opened 2018-02-16
feat(API): Add DBC signal value table 

d-bohls:
Expose Signal property named dbc_signal_value_table
that returns a mapping to a DBC value table.

Add related docs and tests.

### DIFF
--- a/nixnet/database/_dbc_signal_value_table.py
+++ b/nixnet/database/_dbc_signal_value_table.py
@@ -1,0 +1,102 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import collections
+import six
+import typing  # NOQA: F401
+
+from nixnet import _funcs
+from nixnet import constants
+
+
+class DbcSignalValueTable(collections.Mapping):
+    """Collection for accessing a DBC signal value table."""
+
+    def __init__(self, handle):
+        # type: (int) -> None
+        self._handle = handle
+
+    def __repr__(self):
+        return '{}(handle={})'.format(type(self).__name__, self._handle)
+
+    def __eq__(self, other):
+        if isinstance(other, self.__class__):
+            return self._handle == typing.cast(DbcSignalValueTable, other)._handle
+        else:
+            return NotImplemented
+
+    def __ne__(self, other):
+        result = self.__eq__(other)
+        if result is NotImplemented:
+            return result
+        else:
+            return not result
+
+    def __hash__(self):
+        return hash(self._handle)
+
+    def __len__(self):
+        return len(self._value_table)
+
+    def __iter__(self):
+        return self.keys()
+
+    def __getitem__(self, key):
+        # type: (typing.Text) -> typing.Text
+        """Return the value.
+
+            Args:
+                Value description.
+            Returns:
+                Value
+        """
+        if isinstance(key, six.string_types):
+            return self._value_table[key]
+        else:
+            raise TypeError(key)
+
+    def keys(self):
+        """Return all value descriptions in the collection.
+
+            Yields:
+                An iterator to all value descriptions in the collection.
+        """
+        return iter(self._value_table.keys())
+
+    def values(self):
+        """Return all values in the collection.
+
+            Yields:
+                An iterator to all values in the collection.
+        """
+        return iter(self._value_table.values())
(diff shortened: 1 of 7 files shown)


nixnet/database/_dbc_signal_value_table.py
```diff
@@ -0,0 +1,103 @@
...
+
+        # table_string is of the format 'value1, key1, value2, key2, ...'
+        # convert to a dict: { 'key1': int('value1'), 'key2': int('value2'), ... }
+        table_list = [
+            element
+            for element in table_string.split(',')
+            if element.strip()]
+        table_dict = dict(
+            (key, int(value))
+            for value, key in zip(table_list[0::2], table_list[1::2]))
```
jashnani: `if element.strip()` 
Why are we ignoring empty strings when creating a list? Wouldn't this mess with the set of key-value pairings?

Also, why we casting value to int? I didn't see anything about value always being an int. I was thinking we just store the string representation as is (like we do for DBC attributes).

I'm thinking the code would look something like this:

table_list = table_string.split(',')
table_dict = dict(
  (key, value)
  for value, key in zip(table_list[0::2], table_list[1::2])
)

nixnet/database/_dbc_signal_value_table.py
```diff
@@ -0,0 +1,103 @@
...
+            return self._value_table[key]
+        else:
+            raise TypeError(key)
+
+    def keys(self):
+        """Return all value descriptions in the collection.
+
+            Yields:
+                An iterator to all value descriptions in the collection.
+        """
```
jashnani: I wish there was a better name for keys than 'value descriptions'...

tests/test_dbc_signal_value_table.py
```diff
@@ -0,0 +1,71 @@
...
+        assert len({sig1.dbc_signal_value_table, sig2.dbc_signal_value_table}) == 2  # testing `__hash__`
+        assert sig1.dbc_signal_value_table == sig1.dbc_signal_value_table
+        assert sig1.dbc_signal_value_table != sig2.dbc_signal_value_table
+        for key in sig1.dbc_signal_value_table:
+            print(sig1.dbc_signal_value_table[key])
+
+        # test container
+        assert sorted(sig1.dbc_signal_value_table.keys()) == ['High', 'Low', 'Zero']
+        assert sorted(sig1.dbc_signal_value_table.values()) == [-10, 0, 4]
+        assert sorted(sig1.dbc_signal_value_table.items()) == [('High', 4), ('Low', -10), ('Zero', 0)]
```
jashnani: Should we test the case where dbc file returns an odd number of items in the value table string? I think your code will ignore that last item and create key-value pairings for the rest but we should verify it.

nixnet/database/_dbc_signal_value_table.py
d-bohls: Me too.

nixnet/database/_dbc_signal_value_table.py
d-bohls: "Why are we ignoring empty strings when creating a list? Wouldn't this mess with the set of key-value pairings?"

From my experience in playing with DBC creator, I was not concerned about empty keys. I opting for the case where I get `[]` instead of `['']` if `table_string == ''`. I'm happy to guard against empty keys though...I'll make the change.

> Also, why we casting value to int?

I'll send you information from a DBC creation tool that supports value being an int.

tests/test_dbc_signal_value_table.py
d-bohls: My preference is to just raise an exception for that case, but not modify the DBC database with an invalid table.

I'll look into how easy it would be to mock an odd-numbered string list.

nixnet/database/_dbc_signal_value_table.py
d-bohls: Per our conversation, we'll hold off on making value an int unless I can be 100% sure that is all it will ever be.

jashnani: Fixed.

tests/test_dbc_signal_value_table.py
jashnani: Ok. Fixed.

PR #228, opened 2018-02-16
chore(API): increase uniformity of dunders

d-bohls:

### DIFF
--- a/nixnet/database/_cluster.py
+++ b/nixnet/database/_cluster.py
@@ -21,8 +21,9 @@ class Cluster(object):
     """Database cluster"""
 
     def __init__(self, handle):
+        # type: (int) -> None
         self._handle = handle
-        self._dbc_attributes = None
+        self._dbc_attributes = None  # type: typing.Optional[_dbc_attributes.DbcAttributeCollection]
         self._ecus = _collection.DbCollection(
             self._handle, constants.ObjectClass.ECU, _cconsts.NX_PROP_CLST_ECU_REFS, _ecu.Ecu)
         self._frames = _collection.DbCollection(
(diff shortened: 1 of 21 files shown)


nixnet/database/_signal.py
```diff
@@ -13,8 +13,9 @@ class Signal(object):
     """Database signal"""
 
     def __init__(self, handle):
+        # type: (int) -> None
         self._handle = handle
-        self._dbc_attributes = None
+        self._dbc_attributes = None  # type: typing.Union[None, _dbc_attributes.DbcAttributeCollection]
```
epage: You can use `Optional` instead

https://docs.python.org/3/library/typing.html#typing.Optional

nixnet/database/_cluster.py
```diff
@@ -21,8 +21,9 @@ class Cluster(object):
     """Database cluster"""
 
     def __init__(self, handle):
+        # type: (int) -> None
         self._handle = handle
-        self._dbc_attributes = None
+        self._dbc_attributes = None  # type: typing.Union[None, _dbc_attributes.DbcAttributeCollection]
```
d-bohls: Adding a type annotation to this constructor causes mypy to start tracking other variables in init. mypy then began complaining about the type transition of `self._dbc_attributes` from `None` to `_dbc_attributes.DbcAttributeCollection`.

> Incompatible types in assignment (expression has type 'DbcAttributeCollection' variable has type None)

That's why I had to add a type annotation to this line (also done for other classes with DBC attributes) .

nixnet/database/_signal.py
d-bohls: I like that more. I'll do that.

nixnet/database/_cluster.py
epage: btw to make the mypy typing more correct for container functions, I think you can switch from inheriting from `abc.Mapping` (or whatever) to `typing.Mapping[From, To]`.  Not sure if this works for Python2.

Example:
```python
T = TypeVar('T')

class MyDict(Mapping[str, T]):
    ...
```
https://docs.python.org/3/library/typing.html#user-defined-generic-types

PR #233, opened 2018-03-08
refactor(API): have config_status return status code and text.

d-bohls:
BREAKING CHANGE:

Before, we just returned the status code.
However, the user previously had no way to get the text for the status code.
Now, return the status code and text in a tuple.

### DIFF
--- a/nixnet/_utils.py
+++ b/nixnet/_utils.py
@@ -28,8 +28,7 @@ def flatten_items(list):
         # For FRAME_IN_QUEUED / FRAME_OUT_QUEUED
         # Convenience for everything else
         if ',' in list:
-            # A bit of an abuse of an error code
-            _errors.check_for_error(_cconsts.NX_ERR_INVALID_PROPERTY_VALUE)
+            _errors.raise_xnet_error(_cconsts.NX_ERR_INVALID_PROPERTY_VALUE)
         flattened = list
     elif isinstance(list, collections.Iterable):
         flattened = ",".join(list)
@@ -37,8 +36,7 @@ def flatten_items(list):
         # For FRAME_IN_STREAM / FRAME_OUT_STREAM
         flattened = ''
     else:
-        # A bit of an abuse of an error code
-        _errors.check_for_error(_cconsts.NX_ERR_INVALID_PROPERTY_VALUE)
+        _errors.raise_xnet_error(_cconsts.NX_ERR_INVALID_PROPERTY_VALUE)
 
     return flattened
 
--- a/nixnet/_utils.py
+++ b/nixnet/_utils.py
@@ -14,7 +14,7 @@ from nixnet import types
 
 
 def flatten_items(list):
-    # (typing.Union[typing.Text, typing.List[typing.Text]]) -> typing.Text
+    # type: (typing.Union[typing.Text, typing.List[typing.Text]]) -> typing.Text
     """Flatten an item list to a string
 
     >>> str(flatten_items('Item'))
@@ -42,7 +42,7 @@ def flatten_items(list):
 
 
 def parse_can_comm_bitfield(bitfield):
-    # (int) -> types.CanComm
+    # type:  (int) -> types.CanComm
     """Parse a CAN Comm bitfield."""
     state = constants.CanCommState(bitfield & 0x0F)
     tcvr_err = ((bitfield >> 4) & 0x01) != 0
@@ -54,7 +54,7 @@ def parse_can_comm_bitfield(bitfield):
 
 
 def parse_lin_comm_bitfield(first, second):
-    # (int) -> types.CanComm
+    # type: (int, int) -> types.LinComm
     """Parse a LIN Comm first."""
     sleep = ((first >> 1) & 0x01) != 0
     state = constants.LinCommState((first >> 2) & 0x03)
--- a/nixnet/database/_frame.py
+++ b/nixnet/database/_frame.py
@@ -521,8 +521,8 @@ class Frame(object):
         """
         ref = _props.get_frame_mux_data_mux_sig_ref(self._handle)
         if ref == 0:
-            # A bit of an abuse of errors
-            _errors.check_for_error(_cconsts.NX_ERR_SIGNAL_NOT_FOUND)
+            _errors.raise_xnet_error(_cconsts.NX_ERR_SIGNAL_NOT_FOUND)
+
         return _signal.Signal(ref)
 
     @property
--- a/nixnet/database/_frame.py
+++ b/nixnet/database/_frame.py
@@ -47,6 +47,23 @@ class Frame(object):
     def __repr__(self):
         return '{}(handle={})'.format(type(self).__name__, self._handle)
 
+    def check_config_status(self):
+        # type: () -> None
+        """Check this frame's configuration status.
+
+        By default, incorrectly configured frames in the database are not returned from
+        :any:`Cluster.frames` because they cannot be used in the bus communication.
+        You can change this behavior by setting :any:`Database.show_invalid_from_open` to `True`.
+        When a frame configuration status becomes invalid after the database is opened,
+        the frame still is returned from :any:`Cluster.frames`
+        even if :any:`Database.show_invalid_from_open` is `False`.
+
+        Raises:
+            XnetError: The frame is incorrectly configured.
+        """
+        status_code = _props.get_frame_config_status(self._handle)
+        _errors.check_for_error(status_code)
+
     @property
     def application_protocol(self):
         # type: () -> constants.AppProtocol
@@ -81,13 +98,6 @@ class Frame(object):
     def comment(self, value):
         # type: (typing.Text) -> None
         _props.set_frame_comment(self._handle, value)
-
-    @property
-    def config_status(self):
-        # type: () -> typing.Tuple[int, typing.Text]
-        status_code = _props.get_frame_config_status(self._handle)
-        status_text = _errors.status_to_string(status_code)
-        return status_code, status_text
 
     @property
     def default_payload(self):
@@ -572,7 +582,7 @@ class Frame(object):
 
         For CAN and LIN, NI-XNET supports only a one-to-one relationship between frames and PDUs.
         For those interfaces, advanced PDU configuration returns
-        an error from the :any:`Frame.config_status` property and when creating a session.
+        raises an exception when calling :any:`Frame.check_config_status` and when creating a session.
         If you do not use advanced PDU configuration,
         you can avoid using PDUs in the database API
         and create signals and subframes directly on a frame.
--- a/nixnet/database/_frame.py
+++ b/nixnet/database/_frame.py
@@ -84,21 +84,10 @@ class Frame(object):
 
     @property
     def config_status(self):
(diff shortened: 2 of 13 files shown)


epage:
FYI "read state" is another case where status codes are returned.  When I added support for it, I ended up making it a `check_fault` function.

See
- https://github.com/ni/nixnet-python/blob/master/nixnet/_session/base.py#L458
- https://github.com/ni/nixnet-python/pull/131

Should these different cases be consistent, one way or the other?  Or is there a reason for us to take distinct approaches?

d-bohls:
Regarding Ed's comment on "read state":

I talked to numerous people about this and got varying opinions. George H. was of was of the opinion that we should just return Fault from "read state" and not throw an exception because non-zero Fault indicates a network issue; not an XNET/driver issue. I'm thinking I'll probably open a separate issue on what to do with Fault.

d-bohls:
Clearly, what we currently have for config_status is insufficient. We return a status code and no way for the user to call nx_status_to_string to get text for the code.

Here are a few options:

1. Have config_status return a tuple containing the status code and text (this solution I've posted here). For a status code of zero, the string could be empty or none.
2. Handle the status just like other status return values: emit a warning or raise an exception for non-zero status.
3. Maybe we could make status a class that exposes code and text.

Other options? Thoughts?

jashnani:
I could see us going either way on this. `config_status` returning `None` can be somewhat unclear as to whether or not the status is good. For consistency with `check_fault`, we could rename this property to `check_config` or something similar. So we would raise an exception on a bad error code and do nothing otherwise. It'll make the intent behind returning `None` more clear.

trstamper:
Throwing an exception with the status code and error message seems better to me than returning a tuple

d-bohls:
I've looked into this a bit more. If we used tuples, the code might look something like this:

```
        signal_status = signal.config_status
        if signal_status[0] != 0:
            print(signal_status[0])  # print the error code
            print(signal_status[1])  # print the error text
```

If we throw an exception instead, the code looks like this:

```
        try:
            signal.check_config_status()
        except XnetError as e:
            print(e.error_code)  # print the error code
            print(str(e))  # print the error text
```

I ran tests on a database with an invalid signal, and the output for both of these looks like this:

```
3220582728
NI-XNET:  (Hex 0xBFF63148) Frame ConfigStatus: A signal within the frame exceeds the frame boundaries (Payload Length).
```

This exposes a few issues with the code, which are as follows:

1. The C API treats status as an i32. Positive codes are a warning and negative codes are an error. The python API treats status as a u32. It's an error if the MSB is set, and a warning otherwise. If the user is dealing with the code in hex, this is no big deal. However, equivalent error codes as int will not match between the C API and the python API. This doesn't seem ideal to me. Do we care?

2. The issue with unsigned status affects us in other ways too. In _cconsts.py line 8 we define `NX_STATUS_ERROR = (0x80000000)` which comes from the C API and is used as a bit mask for the MSB on an i32 status codes. When we do this in python 2.7, NX_ERROR_BASE get promoted from an int to a long because 3220582400 (the unsigned value of 0x80000000) does fit in a u32. If we set NX_STATUS_ERROR to -2147483648 instead (the i32 value of 0x80000000), this promotion to long doesn't happen. Python 3.6 did away with long so NX_STATUS_ERROR stays an int. So far, the only affect I know of due to this issue is that I ran into difficulty trying to add typing checks on status codes, since codes are a long in 2.7 and an int in 3.6, and 3.6 doesn't know about longs.

3. Question: how should warning codes be handled when calling `check_config_status`? I talked to @sjasonsmith and it may be possible that config status on database objects could return a warning status code. If we send the status code through `_errors.check_for_error`, error codes will raise an exception but warnings will not. Does this match user expectation? Should warning codes also raise an exception in this particular case?

d-bohls:
The last commit was just to show how I would be doing the `check_config_status` command. Disregard the broken CI. Once we settle on the design, I will add a commit to this PR with the correct docstrings and this will fix the CI issue.

jashnani:
> This exposes a few issues with the code, which are as follows:

For # 1 and 2, would it make sense to open a separate issue and track it outside of this PR? I don't think it should be gating the config_status stuff.

For # 3, "how should warning codes be handled when calling check_config_status?" I was wondering the same thing. Maybe we should come up with an API guideline for how to expose check_status or check_whatever type of functionality to the Python API user. I have some ideas here but I can discuss those later after the next release.

For consistency with `check_fault` and for other reasons mentioned in the review comments, for now I think we should just raise an error on a bad error code and return `None` otherwise.

d-bohls:
I filed issue #250 for the previously mentioned status code issues, and modified `check_status_configuration` so that it raises exception or warns. This is ready for review now.

nixnet/_utils.py
```diff
@@ -28,23 +29,25 @@ def flatten_items(list):
         # For FRAME_IN_QUEUED / FRAME_OUT_QUEUED
         # Convenience for everything else
         if ',' in list:
-            # A bit of an abuse of an error code
-            _errors.check_for_error(_cconsts.NX_ERR_INVALID_PROPERTY_VALUE)
+            error_code = _cconsts.NX_ERR_INVALID_PROPERTY_VALUE
+            error_text = _errors.status_to_string(error_code)
+            raise errors.XnetError(error_text, error_code)
```
epage: Why are you no longer calling `check_for_error`?

nixnet/_utils.py
```diff
@@ -28,23 +29,25 @@ def flatten_items(list):
...
         flattened = ",".join(list)
     elif list is None:
         # For FRAME_IN_STREAM / FRAME_OUT_STREAM
         flattened = ''
     else:
-        # A bit of an abuse of an error code
-        _errors.check_for_error(_cconsts.NX_ERR_INVALID_PROPERTY_VALUE)
+        error_code = _cconsts.NX_ERR_INVALID_PROPERTY_VALUE
+        error_text = _errors.status_to_string(error_code)
+        raise errors.XnetError(error_text, error_code)
```
epage: ditto

nixnet/database/_frame.py
```diff
@@ -522,8 +522,9 @@ def mux_data_mux_sig(self):
         """
         ref = _props.get_frame_mux_data_mux_sig_ref(self._handle)
         if ref == 0:
-            # A bit of an abuse of errors
-            _errors.check_for_error(_cconsts.NX_ERR_SIGNAL_NOT_FOUND)
+            error_code = _cconsts.NX_ERR_SIGNAL_NOT_FOUND
+            error_text = _errors.status_to_string(error_code)
+            raise errors.XnetError(error_text, error_code)
```
epage: ditto

nixnet/database/_lin_sched_entry.py
```diff
@@ -53,8 +54,10 @@ def collision_res_sched(self):
         """
         handle = _props.get_lin_sched_entry_collision_res_sched(self._handle)
         if handle == 0:
-            # A bit of an abuse of errors
-            _errors.check_for_error(_cconsts.NX_ERR_DATABASE_OBJECT_NOT_FOUND)
+            error_code = _cconsts.NX_ERR_DATABASE_OBJECT_NOT_FOUND
+            error_text = _errors.status_to_string(error_code)
+            raise errors.XnetError(error_text, error_code)
```
epage: ditto

nixnet/database/_pdu.py
```diff
@@ -202,8 +202,10 @@ def mux_data_mux_sig(self):
         """
         handle = _props.get_pdu_mux_data_mux_sig_ref(self._handle)
         if handle == 0:
-            # A bit of an abuse of errors
-            _errors.check_for_error(_cconsts.NX_ERR_SIGNAL_NOT_FOUND)
+            error_code = _cconsts.NX_ERR_SIGNAL_NOT_FOUND
+            error_text = _errors.status_to_string(error_code)
+            raise errors.XnetError(error_text, error_code)
```
epage: ditto

nixnet/database/_signal.py
```diff
@@ -486,6 +497,8 @@ def mux_subfrm(self):
         from nixnet.database import _subframe
         ref = _props.get_signal_mux_subfrm_ref(self._handle)
         if ref == 0:
-            # A bit of an abuse of errors
-            _errors.check_for_error(_cconsts.NX_ERR_FRAME_NOT_FOUND)
+            error_code = _cconsts.NX_ERR_FRAME_NOT_FOUND
+            error_text = _errors.status_to_string(error_code)
+            raise errors.XnetError(error_text, error_code)
```
epage: ditto

nixnet/_utils.py
epage: I see there is a dedicated commit for this but nothing in the message gives an explanation for it.

d-bohls: Why do you want `check_for_error` to be called? If I know I want to raise an error, why would I call a method that conditionally raises or warns instead of raising the error directly? Originally, I thought this was the "abuse" you were mentioning in the comments, though given your questioning here I suspect you meant the abuse was that you just picked the best fitting error code. That wasn't clear to me.

Side note: in another PR I have, calling `check_for_error` instead of raising directly confuses mypy into complaining that there is a path without a return value.

epage: > Why do you want check_for_error to be called? If I know I want to raise an error, why would I call a method that conditionally raises or warns instead of raising the error directly?

My primary concern is that changes are explained in the commit message.

My (slight) preference for `check_for_error` is that it is a one-stop shop for getting the appropriate behavior in the correct way.  I kept `XnetError`'s `__init__` "low level" for the sake of being easy to create whatever type of `XnetError` we need and assumed everyone would create them via `check_for_error`.

> Originally, I thought this was the "abuse" you were mentioning in the comments, though given your questioning here I suspect you meant the abuse was that you just picked the best fitting error code. That wasn't clear to me.

Yes, it is referring to the error code

> Side note: in another PR I have, calling check_for_error instead of raising directly confuses mypy into complaining that there is a path without a return value.

One way of resolving this is we could add a dependency on `mypy_extensions` and use its [`NoReturn`](https://mypy.readthedocs.io/en/latest/kinds_of_types.html#the-noreturn-type) type annotation.

On the other hand, `check_for_error` is not unconditionally a `NoReturn` due to the warning case.  Hmmm that is one thing in favor of not using `check_for_error`.

d-bohls: What would you think about adding a method in _errors.py

```
def raise_error(error_code):
```

that creates XnetError and raises it. Then, those two methods in _errors.py would be our one-stop shop and my hope is that mypy would be smart enough to figure out that a code path involving `raise_error` unconditionally throws.

d-bohls: I added a new method `raise_xnet_error` to use when we unconditionally want an error, and which is called by `check_for_error`. This also resolve the mypy return issue.

nixnet/_utils.py
```diff
@@ -56,7 +54,7 @@ def parse_can_comm_bitfield(bitfield):
 
 
 def parse_lin_comm_bitfield(first, second):
-    # (int) -> types.CanComm
+    # type: (int, int) -> types.LinComm
```
jashnani: Good catch!

PR #243, opened 2018-03-27
fix(API): DBC signal values are integers.

### LINKED ISSUE #229: Should values of DBC value tables be integers?
Currently, we treat the values of DBC value tables as strings. We have some evidence that they may be restricted to ints. We should verify whether or not the values are always of type integer.

d-bohls:
Fixes #229

BREAKING CHANGE:

DBC signal value is now int instead of string.

### DIFF
--- a/nixnet/database/_dbc_signal_value_table.py
+++ b/nixnet/database/_dbc_signal_value_table.py
@@ -43,7 +43,7 @@ class DbcSignalValueTable(collections.Mapping):
         return self.keys()
 
     def __getitem__(self, key):
-        # type: (typing.Text) -> typing.Text
+        # type: (typing.Text) -> int
         """Return the value.
 
             Args:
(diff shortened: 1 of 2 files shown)


nixnet/database/_dbc_signal_value_table.py
```diff
@@ -97,6 +97,6 @@ def _value_table(self):
         # table_string is of the format 'value1, key1, value2, key2, ...'
         # convert to a dict: { 'key1': int('value1'), 'key2': int('value2'), ... }
         table_dict = dict(
-            (key, value)
+            (key, int(value))
```
jashnani: Based on testing done for #229, we found that this cast is needed because the driver converts value to an int but returns it in a string format (we tested signal values of type string and float).

It may be worth calling it out as a comment.

Issue #53, opened 2017-06-28
`Frame.pdu_*` properties should be considered for a tuple


epage:
#52 adds `db.Frame`.  The pdu properties act as a parallel array.  These should be exposed as a `namedtuple`

d-bohls:
Per the latest C API docs (17.0), it sounds like CAN and LIN are currently limited to a single PDU mapped to a frame.

http://zone.ni.com/reference/en-XX/help/372841P-01/nixnet/nxpropfrm_pdurefs/

Thus, the user's options are:

1. Map a single PDU to a frame.
2. Skip the PDU and add signals and subframes directly to the frame

Only FlexRay allows one or more PDUs to be mapped to one or more frames. We should verify that we can use a PDU with a frame in CAN/LIN, and see if this could possible change in the future. Options are:

1. Return a list of named tuples
2. Return a single named tuple
3. Keep the properties separate and return a single element (pdu, pdu_start_bit, pdu_update_bit)

Something to keep in mind: do we want functionality parity with the database editor? If the database editor allows multi-mapping between PDUs and frames, are how is the customer experience with our python API choices?

lilesj:
CAN and LIN both require the "[PDUs required?](http://zone.ni.com/reference/en-XX/help/372841L-01/nixnet/propertyclusterpdusrequired/)" property to be false. I believe you can still use the database objects to create sessions etc. In LabVIEW we have the Create Session (PDU In/Out Single-Point) family of VIs that use PDUs as inputs and they do work with LIN and CAN so long as "PDUs required?" is false.

d-bohls:
And, to be clear, `PDUs required?` is a read-only boolean property that is `False` when

- Only one PDU is mapped to the frame.
- This PDU is not mapped to other frames.
- The PDU Start Bit in the frame is 0.
- The PDU Update Bit is not used.

So, a very limited case, but it does sound possible for CAN/LIN in the C API as well.

jashnani:
Fixed in #245

PR #245, opened 2018-03-28
`Frame.pdu_*` properties should be considered for a tuple

trstamper:

This PR implements issue #53  by removing the pdu_* properties with a single pdu properties. The old interface required parallel lists, this implementation uses a single list of namedtuples.

#53 

I tested this with my CAN device by creating a PDU that met the requirements of [PDUs Required?](http://zone.ni.com/reference/en-XX/help/372841L-01/nixnet/propertyclusterpdusrequired/). I verified I could configure and take measurements without any warnings. If I didn't conform to PDUs required, I got either a warning or error depending on the case.

### DIFF
--- a/nixnet/database/_frame.py
+++ b/nixnet/database/_frame.py
@@ -2,12 +2,14 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import operator
 import typing  # NOQA: F401
 
 from nixnet import _cconsts
 from nixnet import _errors
 from nixnet import _props
 from nixnet import constants
+from nixnet import types
 
 from nixnet.database import _cluster
 from nixnet.database import _collection
@@ -553,26 +555,20 @@ class Frame(object):
         return self._mux_subframes
 
     @property
-    def pdus(self):
-        # actually returns typing.Iterable[_pdu.Pdu], but avoiding a circular import
-        # type: () -> typing.Iterable[typing.Any]
-        """list of :any:`Pdu`: Get or set a list that maps existing PDUs to a frame.
+    def pdu_properties(self):
+        # type: () -> typing.Iterable[types.PduProperties]
+        """list of :any:`PduProperties`: Get or set a list that maps existing PDUs to a frame.
 
         A mapped PDU is transmitted inside the frame payload when the frame is transmitted.
         You can map one or more PDUs to a frame and one PDU to multiple frames.
 
-        Mapping PDUs to a frame requires setting three frame properties.
-        All three properties are lists of values:
+        Mapping PDUs to a frame requires setting pdu_properties with a list of PduProperties tuples.
+        Each tuple contains the following properties:
 
-        *   :any:`Frame.pdus`: Set this property first to define
-            the sequence of values for the other two properties.
-        *   :any:`Frame.pdu_start_bits`: Defines the start bit of the PDU inside the frame.
-        *   :any:`Frame.pdu_update_bits`: Defines the update bit for the PDU inside the frame.
+        *   :any:`PduProperties.pdu`: Defines the sequence of values for the other two properties.
+        *   :any:`PduProperties.start_bit`: Defines the start bit of the PDU inside the frame.
+        *   :any:`PduProperties.update_bit`: Defines the update bit for the PDU inside the frame.
             If the update bit is not used, set the value to ``-1``.
-
-        Values on the same list position are corresponding.
-        For example, ``pdus[0]``, ``pdu_start_bits[0]``,
-        and ``pdu_update_bits[0]`` define the mapping for the first PDU in the frame.
 
         Databases imported from FIBEX prior to version 3.0,
         from DBC, NCD, or LDF files have a strong one-to-one relationship between frames and PDUs.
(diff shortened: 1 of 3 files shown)


trstamper:
Since CAN/LIN currently only supports a subset of PDU features (one PDU per frame, a start bit of 0, and an update bit of -1), an alternative implementation would be to only have a pdu_ref function which takes a PDU item. We could assume the start bit of 0 and update bit of -1. However, this would be a departure from the XNET C interface, and we'd need to switch back to something like namedtuples if CAN/LIN ever supported more types of PDUs.

nixnet/database/_frame.py
```diff
@@ -239,28 +243,24 @@ def mux_subframes(self):
...
-    @property
-    def pdu_update_bits(self):
-        return _props.get_frame_pdu_update_bits(self._handle)
-
-    @pdu_update_bits.setter
-    def pdu_update_bits(self, value):
-        _props.set_frame_pdu_update_bits(self._handle, value)
+    def pdus(self):
+        handles = [_pdu.Pdu(handle)
+                   for handle in _props.get_frame_pdu_refs(self._handle)]
```
epage: While acceptable by flake8, I personally discourage this style because it leads to unnecessarily deep nesting, and simple refactors (like renaming `handles`) cause  other lines to change (updating indentation)

Alternatively you can do

```python
handles = [
    _pdu.Pdu(handle)
    for handle in _props.get_frame_pdu_refs(self._handle)]
```

nixnet/database/_frame.py
```diff
@@ -239,28 +243,24 @@ def mux_subframes(self):
...
+        pdu_tuples = zip(*(handles,
+                           _props.get_frame_pdu_start_bits(self._handle),
+                           _props.get_frame_pdu_update_bits(self._handle)))
+        for pdu in [types.PduFrameReference(ref, start_bit, update_bit)
+                    for (ref, start_bit, update_bit) in pdu_tuples]:
+            yield pdu
```
epage: Why create a list, then iterate over it, yielding the results?  That feels overly complicated.

Instead you could just
```python
return (
    types.PduFrameReference(ref, start_bit, update_bit)
    for (ref, start_bit, update_bit) in pdu_tuples
)
```
or
```python
for (ref, start_bit, update_bit) in pdu_tuples]:
    yield types.PduFrameReference(ref, start_bit, update_bit)
```

nixnet/database/_frame.py
```diff
@@ -239,28 +243,24 @@ def mux_subframes(self):
...
-        _props.set_frame_pdu_start_bits(self._handle, value)
-
-    @property
-    def pdu_update_bits(self):
-        return _props.get_frame_pdu_update_bits(self._handle)
-
-    @pdu_update_bits.setter
-    def pdu_update_bits(self, value):
-        _props.set_frame_pdu_update_bits(self._handle, value)
+    def pdus(self):
```
epage: Isn't Damon creating documentation for these?  Are you two's commits going to step on each others toes?

nixnet/database/_frame.py
```diff
@@ -83,7 +86,8 @@ def dbc_attributes(self):
         # type: () -> _dbc_attributes.DbcAttributeCollection
         """:any:`nixnet.database._dbc_attributes.DbcAttributeCollection`: Access the frame's DBC attributes."""
         if self._dbc_attributes is None:
-            self._dbc_attributes = _dbc_attributes.DbcAttributeCollection(self._handle)
+            self._dbc_attributes = _dbc_attributes.DbcAttributeCollection(
```
epage: Why is this commit including an unrelated change?

nixnet/database/_frame.py
trstamper: Sure. I was sticking with what autopep8 did, I don't have any horse in this fight.

nixnet/database/_frame.py
trstamper: Yeah, the last one looks cleaner to me too.

nixnet/database/_frame.py
trstamper: Yep. I'll coordinate with damon on how to order these.

nixnet/database/_frame.py
trstamper: I figured I would fix the pep8 violation (over the max line length) while I was in the file. I just looked at tox.ini and saw it's set to 120 characters max, so I'll change this back.

epage: Our official guideline is to aim for below 80 characters but only enforce below 120.

... we still need to get that guideline published publicly.

nixnet/database/_frame.py
epage: I need to get an auto formatter setup.  I hear good things about yapf.  A new one recently came out but I can't find it.

nixnet/database/_frame.py
```diff
@@ -2,15 +2,18 @@
...
 
 from nixnet import _cconsts
 from nixnet import _errors
 from nixnet import _props
 from nixnet import constants
+from nixnet import types
 
 from nixnet.database import _collection
 from nixnet.database import _dbc_attributes
+from nixnet.database import _pdu
```
d-bohls: This import will need to be moved down into the methods that use PDUs, else we will get circular imports when combined with #237.

nixnet/database/_frame.py
```diff
@@ -239,28 +243,24 @@ def mux_subframes(self):
...
+                   for handle in _props.get_frame_pdu_refs(self._handle)]
+        pdu_tuples = zip(*(handles,
+                           _props.get_frame_pdu_start_bits(self._handle),
+                           _props.get_frame_pdu_update_bits(self._handle)))
+        for pdu in [types.PduFrameReference(ref, start_bit, update_bit)
+                    for (ref, start_bit, update_bit) in pdu_tuples]:
+            yield pdu
+
+    @pdus.setter
+    def pdus(self, pdus):
```
d-bohls: #237 changes all ref methods from `object_ref(s)` to `object(s)` and returns the object instead of a ref. Thus, #237 has changed `pdu_refs` to `pdus`. Given that pattern applies to all the other database objects, calling your tuple `pdus` implies the method should return an iterable of PDU objects. Should we perhaps find another name then? Something like `pdu_properties` or `pdu_info`?

Also, you should return a PDU object instead of a handle as well. Just do something like `Pdu(p.ref._handle)`

#237 adds annotations and docstrings for all methods. Moving forward, these will be required for all methods, so you should add these things for your changes as well.

nixnet/types.py
```diff
@@ -153,6 +154,15 @@ class LinComm(LinComm_):
     pass
 
 
+PduFrameReference_ = collections.namedtuple(
+    'PDU_',
+    ['ref', 'start_bit', 'update_bit'])
```
d-bohls: Per my previous comment, this could probably use some renaming since we are moving away from references to objects. My suggestions:

PduFrameReference_ -> PduProperties_
PDU_ -> PduProperties_
ref -> pdu

nixnet/database/_frame.py
d-bohls: @trstamper #237 has been pulled, so whenever you're ready, you'll want to pull down master and merge it with your changes. There will be conflicts to resolve.

nixnet/database/_frame.py
trstamper: I fixed this circular important but ran into another one trying to import nixnet.database._pdu in types.py because the PduProperties class has a Pdu object as one of its values and I wanted to reference it in the documentation. Why are we importing stuff for sphinx in every file instead of adding our whole library to the path in sphinx's conf.py?

d-bohls: > Why are we importing stuff for sphinx in every file instead of adding our whole library to the path in sphinx's conf.py?

I don't follow. What imports are you talking about?

trstamper: Damon and I talked about this offline and I found out I could avoid importing a class for sphinx to see it by not using the fully qualified name.

nixnet/database/_frame.py
```diff
@@ -553,27 +556,21 @@ def mux_subframes(self):
         return self._mux_subframes
 
     @property
-    def pdus(self):
-        # actually returns typing.Iterable[_pdu.Pdu], but avoiding a circular import
-        # type: () -> typing.Iterable[typing.Any]
+    def pdu_properties(self):
+        # type: () -> typing.Iterable[types.PduProperties]
         """list of :any:`Pdu`: Get or set a list that maps existing PDUs to a frame.
```
d-bohls:
```
list of :any:`PduProperties`
```

nixnet/database/_frame.py
```diff
@@ -177,7 +179,8 @@ def dbc_attributes(self):
         # type: () -> _dbc_attributes.DbcAttributeCollection
         """:any:`DbcAttributeCollection`: Access the frame's DBC attributes."""
         if self._dbc_attributes is None:
-            self._dbc_attributes = _dbc_attributes.DbcAttributeCollection(self._handle)
+            self._dbc_attributes = _dbc_attributes.DbcAttributeCollection(
```
d-bohls: No need to make this change.

nixnet/database/_frame.py
```diff
@@ -553,27 +556,21 @@ def mux_subframes(self):
...
 
         Mapping PDUs to a frame requires setting three frame properties.
-        All three properties are lists of values:
+        All three properties are combined into a PduProperties tuple.
 
-        *   :any:`Frame.pdus`: Set this property first to define
-            the sequence of values for the other two properties.
-        *   :any:`Frame.pdu_start_bits`: Defines the start bit of the PDU inside the frame.
-        *   :any:`Frame.pdu_update_bits`: Defines the update bit for the PDU inside the frame.
+        *   :any:`nixnet.types.PduProperties.ref`: Defines the sequence of values for the other two properties.
```
d-bohls: For consistency with other db docs
```
:any:`PduProperties.pdu`
:any:`PduProperties.start_bit`
:any:`PduProperties.update_bit`
```

Rename ref to pdu, per my comment later in the review.

nixnet/types.py
```diff
@@ -153,6 +154,24 @@ class LinComm(LinComm_):
     pass
 
 
+PduProperties_ = collections.namedtuple(
+    'PDU_',
```
d-bohls: Seems like the typename of the named tuple should be something like PDU_PROPERTIES_ instead of PDU_. Not sure if this naming really matters.

We use handle/ref to talk about an integer that the driver uses for the object. We are dealing with an actual Pdu object here; not a ref. Change to:

```
['pdu', 'start_bit', 'update_bit'])
```

nixnet/database/_frame.py
```diff
@@ -553,27 +556,21 @@ def mux_subframes(self):
...
-        # actually returns typing.Iterable[_pdu.Pdu], but avoiding a circular import
-        # type: () -> typing.Iterable[typing.Any]
+    def pdu_properties(self):
+        # type: () -> typing.Iterable[types.PduProperties]
         """list of :any:`Pdu`: Get or set a list that maps existing PDUs to a frame.
 
         A mapped PDU is transmitted inside the frame payload when the frame is transmitted.
         You can map one or more PDUs to a frame and one PDU to multiple frames.
 
         Mapping PDUs to a frame requires setting three frame properties.
```
d-bohls: This isn't true anymore. Maybe something like:

```
Mapping PDUs to a frame requires modifying this list of :any:`PduProperties`,
each of which requires the following three properties:
```

nixnet/database/_frame.py
```diff
@@ -589,73 +586,22 @@ def pdus(self):
...
-        For example, ``pdus[0]``, ``pdu_start_bits[0]``,
-        and ``pdu_update_bits[0]`` define the mapping for the first PDU in the frame.
-        """
-        return _props.get_frame_pdu_update_bits(self._handle)
-
-    @pdu_update_bits.setter
-    def pdu_update_bits(self, value):
-        # type: (typing.List[int]) -> None
-        _props.set_frame_pdu_update_bits(self._handle, value)
+        handles = [_pdu.Pdu(handle) for handle in _props.get_frame_pdu_refs(self._handle)]
```
d-bohls:
```
pdus = [_pdu.Pdu(handle) for handle in ...
```

The handles have been converted to pdu objects.

nixnet/database/_frame.py
trstamper: Yeah whoops, must have missed that this got redone in the merge

nixnet/database/_frame.py
trstamper: Yeahthat makes more sense. I was referring to the underlying driver calls here but that's not necessary at this spot.

trstamper:
Updated review to address comments.

trstamper:
Updated description to add BREKAING CHANGE tag

jashnani:
For future reference, you can add `Fixes #issue` in the commit description and it will automatically close the issue when the commit is pulled into mainline.

Issue #238, opened 2018-03-21
Add tests for database objects


d-bohls:
Add unit tests and integrations tests for everything in the database folder.

PR #248, opened 2018-04-11
feat(API): Add descendant find for parent db objects.

### LINKED ISSUE #239: Exposing descendant find
Exposing descendant find (like [myFrameA.mySignal](http://zone.ni.com/reference/en-XX/help/372841N-01/nixnet/nxdbfindobject/)). Via `getitem`, you can get a direct
descendant by name but not indirect descendants. Each parent object
will need an explicit `find` added to it. This will also run into
problems with circular imports.

d-bohls:
Fixes #239

The following parent database objects now have a find method:
- database
- cluster
- lin schedule
- pdu
- frame
- subframe

All database object are now subclasses of
DatabaseObject, which is used as an interface.
Database objects are now exposed through the
nixnet.database namespace, and their constructors
now take kwargs to signal to users that these
objects should not be instantiated by users.

I tested on the shipping example database "NIXNET_example"
to ensure I can indeed find children and grandchildren,
and that sane errors are thrown when the object name
not found or the object class is wrong.
Unit tests will be added in #238.

- [ ] ~~New tests have been created for any new features or regression tests for bugfixes.~~

### DIFF
--- a/nixnet/database/__init__.py
+++ b/nixnet/database/__init__.py
@@ -3,7 +3,26 @@ from __future__ import division
 from __future__ import print_function
 
 
+from nixnet.database._cluster import Cluster
+from nixnet.database._database_object import DatabaseObject
+from nixnet.database._ecu import Ecu
+from nixnet.database._frame import Frame
+from nixnet.database._lin_sched import LinSched
+from nixnet.database._lin_sched_entry import LinSchedEntry
+from nixnet.database._pdu import Pdu
+from nixnet.database._signal import Signal
+from nixnet.database._subframe import SubFrame
 from nixnet.database.database import Database
 
 
-__all__ = ["Database"]
+__all__ = [
+    "Cluster",
+    "Database",
+    "DatabaseObject",
+    "Ecu",
+    "Frame",
+    "LinSched",
+    "LinSchedEntry",
+    "Pdu",
+    "Signal",
+    "SubFrame"]
--- a/nixnet/database/_collection.py
+++ b/nixnet/database/_collection.py
@@ -11,6 +11,8 @@ from nixnet import _cprops
 from nixnet import _funcs
 from nixnet import constants  # NOQA: F401
 
+from nixnet.database import _database_object  # NOQA: F401
+
 
 class DbCollection(collections.Mapping):
     """Collection of database objects."""
@@ -58,7 +60,7 @@ class DbCollection(collections.Mapping):
         """
         if isinstance(index, six.string_types):
             ref = _funcs.nxdb_find_object(self._handle, self._type, index)
-            return self._factory(ref)
+            return self._factory(_handle=ref)
         else:
             raise TypeError(index)
 
@@ -93,15 +95,17 @@ class DbCollection(collections.Mapping):
             yield child.name, child
 
     def add(self, name):
-        # type: (typing.Text) -> typing.Any
+        # type: (typing.Text) -> _database_object.DatabaseObject
         """Add a new database object to the collection.
 
         Args:
             name(str): Name of the new database object.
+        Returns:
+            ``DatabaseObject``: An instance of the new database object.
         """
         ref = _funcs.nxdb_create_object(self._handle, self._type, name)
-        return self._factory(ref)
+        return self._factory(_handle=ref)
 
     def _get_children(self):
         for ref in _cprops.get_database_ref_array(self._handle, self._prop_id):
-            yield self._factory(ref)
+            yield self._factory(_handle=ref)
--- a/nixnet/database/_database_object.py
+++ b/nixnet/database/_database_object.py
@@ -1,0 +1,7 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+
+class DatabaseObject(object):
+    """Database object interface."""
--- a/nixnet/database/_find_object.py
+++ b/nixnet/database/_find_object.py
@@ -1,0 +1,47 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import typing  # NOQA: F401
+
+from nixnet import _cconsts
+from nixnet import _errors
+from nixnet import _funcs
+from nixnet import constants
+from nixnet.database import _database_object  # NOQA: F401
+
+
+def find_object(
+        parent_handle,  # type: int
+        object_class,  # type: typing.Any
+        object_name,  # type: typing.Text
+):
+    # type: (...) -> _database_object.DatabaseObject
+    from nixnet.database._cluster import Cluster
+    from nixnet.database._ecu import Ecu
+    from nixnet.database._frame import Frame
+    from nixnet.database._lin_sched import LinSched
+    from nixnet.database._lin_sched_entry import LinSchedEntry
+    from nixnet.database._pdu import Pdu
+    from nixnet.database._signal import Signal
+    from nixnet.database._subframe import SubFrame
+
+    class_enum = {
+        Cluster: constants.ObjectClass.CLUSTER,
+        Ecu: constants.ObjectClass.ECU,
+        Frame: constants.ObjectClass.FRAME,
+        LinSched: constants.ObjectClass.LIN_SCHED,
+        LinSchedEntry: constants.ObjectClass.LIN_SCHED_ENTRY,
+        Pdu: constants.ObjectClass.PDU,
+        Signal: constants.ObjectClass.SIGNAL,
+        SubFrame: constants.ObjectClass.SUBFRAME,
+    }.get(object_class)
+
+    if class_enum is None:
+        raise ValueError("Unsupported value provided for argument object_class.", object_class)
+
+    found_handle = _funcs.nxdb_find_object(parent_handle, class_enum, object_name)
+    if found_handle == 0:
+        _errors.raise_xnet_error(_cconsts.NX_ERR_DATABASE_OBJECT_NOT_FOUND)
+
+    return object_class(_handle=found_handle)
--- a/nixnet/database/_signal.py
+++ b/nixnet/database/_signal.py
@@ -9,16 +9,30 @@ from nixnet import _errors
 from nixnet import _props
 from nixnet import constants
 
+from nixnet.database import _database_object
 from nixnet.database import _dbc_attributes
 from nixnet.database import _dbc_signal_value_table
 
+# workaround to avoid circular imports caused by mypy type annotations
+MYPY = False
+if MYPY:
+    from nixnet.database import _frame  # NOQA: F401
+    from nixnet.database import _pdu  # NOQA: F401
+    from nixnet.database import _subframe  # NOQA: F401
 
-class Signal(object):
+
+class Signal(_database_object.DatabaseObject):
     """Database signal"""
 
-    def __init__(self, handle):
-        # type: (int) -> None
-        self._handle = handle
+    def __init__(
+            self,
+            **kwargs  # type: int
+    ):
+        # type: (...) -> None
+        if not kwargs or '_handle' not in kwargs:
+            raise TypeError()
+
+        self._handle = kwargs['_handle']
         self._dbc_attributes = None  # type: typing.Optional[_dbc_attributes.DbcAttributeCollection]
         self._dbc_signal_value_table = _dbc_signal_value_table.DbcSignalValueTable(self._handle)
 
@@ -64,7 +78,7 @@ class Signal(object):
             (for example, a CAN frame is defined with more than 8 payload bytes).
 
         Raises:
-            XnetError: The signal is incorrectly configured.
+            :any:`XnetError`: The signal is incorrectly configured.
         """
         status_code = _props.get_signal_config_status(self._handle)
         _errors.check_for_error(status_code)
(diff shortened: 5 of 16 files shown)


d-bohls:
A couple of notes on this review.

After #237 goes in, I plan to add docstrings to this review, because the docstrings for this review will reference docstrings in the other review.

I am deferring unit tests for this review to be done as part of #238.

epage:
Would it be possible to post a comment of what your modified example looks like, so we can see what the API is like to use?  Or more ideally, tests :)

nixnet/database/_find_object.py
```diff
@@ -0,0 +1,52 @@
...
+        return LinSched(found_handle)
+    elif object_class == constants.ObjectClass.LIN_SCHED_ENTRY:
+        from nixnet.database._lin_sched_entry import LinSchedEntry
+        return LinSchedEntry(found_handle)
+    else:
+        # A bit of an abuse of an error code
+        _errors.check_for_error(_cconsts.NX_ERR_INVALID_PROPERTY_VALUE)
+        # The above line will raise an exception.
+        # However, mypy doesn't know that and will complain if the following return is missing.
+        return None
```
epage: Looks like you could make the return type a union of the regular and [`NoReturn`](https://mypy.readthedocs.io/en/latest/kinds_of_types.html#the-noreturn-type)

nixnet/database/_find_object.py
```diff
@@ -0,0 +1,52 @@
...
+import typing  # NOQA: F401
+
+from nixnet import _cconsts
+from nixnet import _errors
+from nixnet import _funcs
+from nixnet import constants
+
+
+def find_object(parent_handle, object_class, object_name):
+    # type: (int, constants.ObjectClass, typing.Text) -> object
```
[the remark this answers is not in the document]
d-bohls: Regarding 1, I had thought about doing it that way. I guess on the plus side, it eliminates an enum the user has to deal with, and encourages familiarity with the actual objects. On the con side, those of us who use IDEs lose the benefit of intellisense for the enum values. I think my preference would be to continue using the enum as I don't really see much benefit to passing in an object instead of an enum. I'm open to either way though.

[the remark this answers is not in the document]
epage: > On the con side, those of us who use IDEs lose the benefit of intellisense for the enum values. 

Completion for `nixnet.constants.ObjectClass` is not much different than completion for `nixnet.database.`

As for other intellisense features, you can improve that by
1. Providing a common base class for all database objects
2. Change the annotations from `type: (int, constants.ObjectClass, typing.Text) -> object` to `type: (int, typing.Type[nixnet.database.SomeBaseClass], typing.Text) -> nixnet.database.SomeBaseClass`

This is similar to the approach we took elsewhere in the API
- Define [`FrameFactory`](https://github.com/ni/nixnet-python/blob/master/nixnet/types.py#L248) have have each frame implement it
- Accept a frame object when [reading](https://github.com/ni/nixnet-python/blob/7f79f5b3a23c5ffda8f965fcd2d03ad2316f8b88/nixnet/_session/frames.py#L97) and [converting](https://github.com/ni/nixnet-python/blob/e9a946b035db388030b13c426a58f01a5afe4de2/nixnet/convert.py#L216)

> I think my preference would be to continue using the enum as I don't really see much benefit to passing in an object instead of an enum.

Advantages
- Minimize vocabulary terms the user of the API has to deal with.  I need a `Cluster`, so let me pass a `Cluster` vs having to deal with an enum that represents `Cluster`.
- Avoid C-isms.  Part of the role of the enum is a form of RTTI in C.  Changing it to the actual python class is exposing the same concept in the python way
- Controlled access.  The doc string and intellisense will help guide people to the correct values to pass in rather than having things like `System` be an option.

epage:
For reference, with using classes instead of enums

```python
import nixnet
from nixnet import database


def main():
    with database.Database("NIXNET_example") as db:
        cluster = db.find(database.Cluster, "CAN_Cluster")
        print(cluster.name)
        frame1 = db.find(database.Frame, "CANCyclicFrame2")
        frame2 = cluster.find(database.Frame, "CANCyclicFrame2")
        print(frame1.name)
        print(frame1.name)
        assert (frame1 == frame2)
        signal = db.find(database.Signal, "ClutchPressure")
        print(signal.name)
```

Oh, fun, that won't work as of this moment.  `nixnet.database` only exposes `Database` and not the other types
https://github.com/ni/nixnet-python/blob/master/nixnet/database/__init__.py

Interesting.  I know the concern was about exposing the useless private constructors to users but this doesn't allow our users to use mypy.

A workaround for exposing the useless private constructors is to make `handle` a keyword-only arg that is prefixed with `_`.  That isn't officially supported in python2 but I recently learned about a backport of the feature: https://pypi.python.org/pypi/kwonly-args

d-bohls:
I'm curious about your reasoning for suggesting kwonly-args and not kwargs? Won't both work fine for this purpose? kwonly-args looks like it might be a cleaner implementation.

epage:
> I'm curious about your reasoning for suggesting kwonly-args and not kwargs? Won't both work fine for this purpose? kwonly-args looks like it might be a cleaner implementation.

If you are taking about usage, any keyword argument can also be a positional argument.

If you are talking about `**kwargs`, then that is effectively another way of defining a keyword-only argument and fits the spirit of what I'm suggesting.  The literal keyword-only argument feature just requires less boiler plate / is more clear on intent.

d-bohls:
I've done the following:

1. Introduced DatabaseObjectBase for all database object to inherit from
2. Exposed them in database package's init
3. Modified find methods to use DatabaseObjectBase objects

However, I'm still not seeing a useful namespacing experience. Not sure yet what I'm missing.

I'm holding off on producing docstrings and making the kwargs constructor changes until I can fix the namespacing experience.

nixnet/database/_find_object.py
```diff
@@ -26,28 +26,31 @@ def find_object(
...
+        class_enum = constants.ObjectClass.PDU
+    elif isinstance(object_class, Signal):
+        class_enum = constants.ObjectClass.PDU
+    elif isinstance(object_class, Signal):
+        class_enum = constants.ObjectClass.SIGNAL
+    elif isinstance(object_class, SubFrame):
+        class_enum = constants.ObjectClass.SUBFRAME
+    else:
+        # arbitrary value for class_enum so mypy sees a value in each condition
+        class_enum = constants.ObjectClass.CLUSTER
```
epage: I'd actually recommend a `raise ValueError("Unsupported database object")`

epage: Oh, missed the raise below :)

nixnet/database/__init__.py
```diff
@@ -3,7 +3,24 @@
...
+from nixnet.database._lin_sched import LinSched
+from nixnet.database._lin_sched_entry import LinSchedEntry
+from nixnet.database._pdu import Pdu
+from nixnet.database._signal import Signal
+from nixnet.database._subframe import SubFrame
 from nixnet.database.database import Database
 
 
-__all__ = ["Database"]
+__all__ = [
```
epage: `DatabaseObjectBase` should be in here so users can do mypy typing

nixnet/database/_database_object_base.py
```diff
@@ -0,0 +1,31 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import typing  # NOQA: F401
+
+
+class DatabaseObjectBase(object):
```
epage: Since this should be public, we should ensure we have a name we like.  I personally find -`Base` suffixes a little off, at least in python code.

epage:
> I've done the following:

Looks slick

> However, I'm still not seeing a useful namespacing experience. Not sure yet what I'm missing.

What do you mean?

nixnet/database/_database_object_base.py
d-bohls: I was following what I saw with [SessionBase](https://github.com/ni/nixnet-python/blob/master/nixnet/_session/base.py#L22). So your suggestion is `DatabaseObject`?

epage: iirc SessionBase was me doing what I said not to do, concrete inheritance for code sharing.  Its not actually exposed in the API.

Something like `DatabaseObject` might work.

d-bohls:
@epage 

> However, I'm still not seeing a useful namespacing experience. Not sure yet what I'm missing.
> What do you mean?

Apologies for not be clear there. I was talking about the intellisense issue where I don't see anything useful in the database namespace after typing "database.".

Turns out, this seems to be a PyCharm quirk where my nixnet-python package was defined in my PyCharm project (rather than being installed traditionally) and using a loose python script outside of the project to test with. It's strange because the script would run just fine, but I wasn't getting any intellisense for things in nixnet. Once I include my test files as a source root in the project, the intellisense automagically works.

nixnet/database/__init__.py
d-bohls: done

d-bohls:
Added docstrings and modified database objects to init with kwargs.

@jashnani and I chose to not include docs for the new DatabaseObject base class.

Thanks to @jashnani and @epage for not settling on "good enough". I really like the final design.

I'll add tests for this and all database objects next.

d-bohls:
@jashnani Just in case, let me squash before merge.

d-bohls:
The only "less than ideal" issue I might mention is that the `DatabaseObject` base class shows up in intellisense, though this is better than the enum alone since other things like `Session` showed up in that.

image

nixnet/database/_database_object.py
```diff
@@ -0,0 +1,41 @@
...
+        if result is NotImplemented:
+            return result
+        else:
+            return not result
+
+    def __hash__(self):
+        return hash(self._handle)
+
+    def __repr__(self):
+        return '{}(handle={})'.format(type(self).__name__, self._handle)
```
jashnani: Our use of DatabaseObject is meant for interface only. I would've expected these implementations to continue living in child classes and evolve as necessary:

`__eq__`
`__ne__`
`__hash__`
`__repr__`

nixnet/database/_signal.py
```diff
@@ -483,19 +468,19 @@ def mux_value(self):
     @property
     def mux_subfrm(self):
         # actually returns _subframe.SubFrame, but avoiding a circular import
-        # type: () -> typing.Any
+        # type: () -> _database_object.DatabaseObject
```
jashnani: Should the return type just be SubFrame? Ditto as above.

nixnet/database/_signal.py
```diff
@@ -187,14 +172,14 @@ def default(self, value):
     @property
     def frame(self):
         # actually returns _frame.Frame, but avoiding a circular import
-        # type: () -> typing.Any
+        # type: () -> _database_object.DatabaseObject
```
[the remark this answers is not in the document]
d-bohls: > Shouldn't the return type just be Frame?

Yes, that's what the comment is saying. I couldn't find a way to import before the type check, and not for lack of trying.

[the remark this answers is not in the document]
jashnani: There's a way to do it by setting MYPY = false. See details here: https://mypy.readthedocs.io/en/latest/common_issues.html#import-cycles

I'm not going to hold the review for this, so don't feel like you have to go out of way to do this.

nixnet/database/_database_object.py
```diff
@@ -0,0 +1,41 @@
...
+class DatabaseObject(object):
+    """Database object base."""
+
+    def __init__(
+            self,
+            **kwargs  # type: int
+    ):
+        # type: (...) -> None
+        if not kwargs or '_handle' not in kwargs:
+            _errors.raise_xnet_error(_cconsts.NX_ERR_INVALID_PROPERTY_VALUE)
```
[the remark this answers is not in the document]
d-bohls: You mentioned this PEP for keyword-only arguments:

https://www.python.org/dev/peps/pep-3102/ 

"In accordance with the current Python implementation, any errors encountered will be signaled by raising TypeError."

Changed to raise TypeError(),

nixnet/database/_find_object.py
```diff
@@ -0,0 +1,54 @@
...
+        class_enum = constants.ObjectClass.LIN_SCHED
+    elif object_class is LinSchedEntry:
+        class_enum = constants.ObjectClass.LIN_SCHED_ENTRY
+    elif object_class is Pdu:
+        class_enum = constants.ObjectClass.PDU
+    elif object_class is Signal:
+        class_enum = constants.ObjectClass.SIGNAL
+    elif object_class is SubFrame:
+        class_enum = constants.ObjectClass.SUBFRAME
+    else:
```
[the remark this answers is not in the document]
d-bohls: I had thought about using a dictionary but had no strong preference.

It's a dictionary now. If the key isn't found, I'm raising a ValueError for the object_class argument.

nixnet/database/_collection.py
```diff
@@ -58,7 +58,7 @@ def __getitem__(self, index):
         """
         if isinstance(index, six.string_types):
             ref = _funcs.nxdb_find_object(self._handle, self._type, index)
-            return self._factory(ref)
+            return self._factory(_handle=ref)
```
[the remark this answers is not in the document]
d-bohls: Done

nixnet/database/_signal.py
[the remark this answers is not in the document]
d-bohls: I was able to get this to work using `MYPY = false`. Cool.

nixnet/database/_signal.py
d-bohls: I was able to get this to work using the `MYPY = False` technique.

nixnet/database/_subframe.py
```diff
@@ -141,7 +177,7 @@ def name(self, value):
     @property
     def pdu(self):
         # actually returns _pdu.Pdu, but avoiding a circular import
-        # type: () -> typing.Any
+        # type: () -> _database_object.DatabaseObject
```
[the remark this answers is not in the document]
d-bohls: I was able to get this to work using the `MYPY = False` technique.

nixnet/database/_database_object.py
d-bohls: I read the information you sent me on this. I do understand some of the benefits mentioned to not DNR (like code locality). In this particular situation, I think I'd still prefer not to duplicate theses four functions in all eight subclasses. Regardless, I've moved them back as they were and I'll add unit tests as appropriate in my upcoming PR.

PR #256, opened 2018-05-02
 fix(API): Use _ctypedefs.char_p to send a string to the driver.

d-bohls:
I believe we were copying the string into a buffer
without null termination, so the C API stores a longer
string including non-Ascii characters. Then, when we get
that string back from the driver, we get the non-Ascii
characters too, and then get an error when trying to
decode into an Ascii string.

- [X] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nixnet-python/blob/master/CONTRIBUTING.rst).
- [ ] ~~New tests have been created for any new features or regression tests for bugfixes.~~


epage:
- What issue did you observe that this is fixing?
- Why do you feel switching from `create_string_buffer` to `char_p` fixes this?
- What did you observe that makes you think this fixes it?

From reading the docs, `char_p` doesn't guarentee a null terminator but create_string_buffer says this:

> If a bytes object is specified as first argument, the buffer is made one item larger than its length so that the last element in the array is a NUL termination character. An integer can be passed as second argument which allows specifying the size of the array if the length of the bytes should not be used.

d-bohls:
From your comment, I'm reading that an alternative solution may be to remove the length argument from `create_string_buffer`.

d-bohls:
I'm having trouble finding anything official on how to pass a null terminated string to a `char*`.

I found a few non-official sources suggesting to use `_ctypedefs.char_p(str.encode('ascii'))` and other sources saying the safest bet to pass a null-terminated string is to do ctypes.create_string_buffer(str.encode("ascii")).

I confirmed it also fixes the bug if I leave the length argument out of `create_string_buffer`,

@jashnani, what do you want to do here? I get the impression that the "safest" way forward would be to use `create_string_buffer` instead of `char_p` when passing strings to `char*`. Do we want to change this everywhere?

Issue #255, opened 2018-04-27
Expose new API features from NI-XNET 17.0 and 17.5


jashnani:
Add support for any new API elements introduced NI-XNET 17.0 and 17.5

d-bohls:
I diffed nixnet.h 16.1 against 18.5 and came up with this list of new items:

errors:
                nxErrLowLevelCommunicationTimeout
                nxErrCdaqTransferAborted
                nxErrDatabaseImportVersion
                nxErrEventUnequalPayloadLength
                nxErrEventUnequalCheckSumType
                nxErrCrioBadDriverVersions
                nxErrCrioMissingSlotSupport
                nxErrIntfAlreadyInUseByNiXnet
                nxErrIntfAlreadyInUseByNiXcl
                nxErrByteArrayNotAllowed
                nxErrNoByteArrayMix
                nxErrOnlyOneByteArray

properties:
                nxPropSession_IntfLINChecksumToInStrm
                nxPropSession_J1939IncludeDestAddrInPGN

enum values:
                nxDevForm_PCIe
                nxSigDataType_ByteArray

methods:
                nxConvertFramesToByteArraySinglePoint
                nxConvertByteArrayToFramesSinglePoint

PR #263, opened 2018-08-20
feat(API): Add properties available in NI-XNET 18.0.

d-bohls:

This change adds the new properties listed in #255.
This change does not address the two new methods for
converting byte arrays.

Add a new enum value:
* DevForm.PCIE

Add new error codes:
* NX_ERR_LOW_LEVEL_COMMUNICATION_TIMEOUT
* NX_ERR_CDAQ_TRANSFER_ABORTED
* NX_ERR_DATABASE_IMPORT_VERSION
* NX_ERR_EVENT_UNEQUAL_PAYLOAD_LENGTH
* NX_ERR_EVENT_UNEQUAL_CHECKSUM_TYPE
* NX_ERR_CRIO_BAD_DRIVER_VERSIONS
* NX_ERR_CRIO_MISSING_SLOT_SUPPORT
* NX_ERR_INTF_ALREADY_IN_USE_BY_NI_XNET
* NX_ERR_INTF_ALREADY_IN_USE_BY_NI_XCL
* NX_ERR_BYTE_ARRAY_NOT_ALLOWED
* NX_ERR_NO_BYTE_ARRAY_MIX
* NX_ERR_ONLY_ONE_BYTE_ARRAY

Sort existing errors (matches changes made in nixnet.h):
* NX_ERR_SESSION_TYPE_FRAME_INCOMPATIBILITY
* NX_ERR_DIAGNOSTIC_SCHEDULE_NOT_DEFINED

Add two new driver properties, along with unit tests for them:
* Interface.lin_checksum_to_in_strm
* J1939.include_dest_addr_in_pgn

Here is the equivalent C API documentation for the new properties:

* http://zone.ni.com/reference/en-XX/help/372841T-01/nixnet/nxpropsession_intflinchecksumtoinstrm/
* http://zone.ni.com/reference/en-XX/help/372841T-01/nixnet/nxpropsession_j1939includedestaddrinpgn/


nixnet/_cconsts.py
```diff
@@ -105,7 +105,6 @@
 NX_ERR_FRAME_SIZE_MISMATCH = (NX_ERROR_BASE | 0x0AA)
 NX_ERR_INDEX_TOO_BIG = (NX_ERROR_BASE | 0x0AB)
 NX_ERR_SESSION_MODE_INCOMPATIBILITY = (NX_ERROR_BASE | 0x0AC)
-NX_ERR_SESSION_TYPE_FRAME_INCOMPATIBILITY = (NX_ERROR_BASE | 0x15D)
```
epage: These were moved and not removed

tests/test_j1939.py
```diff
@@ -0,0 +1,21 @@
...
+
+MockXnetLibrary = mock.create_autospec(_cfuncs.XnetLibrary, spec_set=True, instance=True)
+MockXnetLibrary.nx_get_property.return_value = _ctypedefs.u32(0)
+MockXnetLibrary.nx_set_property.return_value = _ctypedefs.u32(0)
+
+
+@mock.patch('nixnet._cfuncs.lib', MockXnetLibrary)
+def test_j1939_properties():
+    # This test should be expanded to cover more J1939 properties.
+    j1939.J1939.include_dest_addr_in_pgn = j1939.J1939.include_dest_addr_in_pgn
```
epage: My main concern with this PR is the testing strategy.

This calls the getter and setter on the upper layers but ends in a dead end at the `_cfuncs.XnetLibrary` layer.  What value does this offer us over mypy's checks?

nixnet/_cconsts.py
d-bohls: Yup, a change to nixnet.h sorted them (by error code) and so I did too, as mentioned in the commit.

tests/test_j1939.py
d-bohls: For methods/properties which are simply thinly layered wrappers to a library call, I've questioned the value of using unit tests over integration tests. I've advocated that we test them all in integration tests since you get the benefit of making sure that (a) the driver can correctly handle the arguments from python and that (b) python can correctly handle an actual return value from the driver. I was given the guidance that we should do some percentage of properties/methods (was it 10 or 15%, I can't remember) as integration tests and then use unit tests for the others. I have attempted to follow that guidance. It would be nice going forward if we had a testing strategy that addresses these issues.

nixnet/_enums.py
```diff
@@ -1674,12 +1674,22 @@ class Phase(enum.Enum):
...
+        PXIE
+        USB
+        PCIE
+    '''
     PXI = _cconsts.NX_DEV_FORM_PXI
     PCI = _cconsts.NX_DEV_FORM_PCI
     C_SERIES = _cconsts.NX_DEV_FORM_C_SERIES
     PXIE = _cconsts.NX_DEV_FORM_PXIE
     USB = _cconsts.NX_DEV_FORM_USB
+    PCIE = _cconsts.NX_DEV_FORM_PCIE
```
jashnani: Very minor. it might help if we sorted the values by bus type or alphabetically (both in code and in docstring) from legibility standpoint.

For example:

C_SERIES
PCI
PCIE
PXI
PXIE
USB

Super minor so not going to hold the review for this.

d-bohls: I agree; sorted is better and there is no need to keep these values in the same order as the C API. Done.

d-bohls:
After I changed the tests from unit to integration tests, the integration tests started failing on Jenkins because the test machine has an older XNET driver that doesn't support the newer properties. I'll upgrade the XNET driver on the test machine and run the tests again.

PR #306, opened 2025-04-16
Fix flake8 W503 error and remove H903 from flake8 ignore list

olsl21:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nixnet-python/blob/main/CONTRIBUTING.rst).
~- [ ] New tests have been created for any new features or regression tests for bugfixes.~
- [x] `tox` successfully runs, including unit tests and style checks (see [CONTRIBUTING.md](https://github.com/ni/nixnet-python/blob/main/CONTRIBUTING.rst)).

This PR is to fix the flake8 W503 (Line break occurred before a binary operator) error that will be thrown when the `flake8` dependency is upgraded to `v7.1.2` to support newer Python versions.

`W503` conflicts with `W504` (Line break occurred after a binary operator), hence we'll need to disable one of them. This PR adds `W503` to the flake8 ignore list in tox.ini.

This PR also removes `H903: Windows style line endings not allowed in code` from the existing flake8 ignore list since [this error code does not seem to be valid](https://www.flake8rules.com/).

This PR is part of a series of PRs to prepare for supporting newer Python versions.

- Locally run `tox` successfully with Python 3.7.9
- Locally run `tox` successfully with Python 3.11.9 with locally upgraded dependencies

### DIFF
--- a/nixnet/_session/collection.py
+++ b/nixnet/_session/collection.py
@@ -85,7 +85,7 @@ class Collection(Sequence):
             other_collection = typing.cast(Collection, other)
             return (
                 self._handle == other_collection._handle
-                and self._list_cache == other_collection._list_cache
+                and self._list_cache == other_collection._list_cache  # NOQA: W503
             )
         else:
             return NotImplemented
@@ -127,7 +127,7 @@ class Item(object):
             other_item = typing.cast(Item, other)
             return (
                 self._handle == other_item._handle
-                and self._index == other_item._index
+                and self._index == other_item._index  # NOQA: W503
             )
         else:
             return NotImplemented
--- a/nixnet/_session/collection.py
+++ b/nixnet/_session/collection.py
@@ -85,7 +85,7 @@ class Collection(Sequence):
             other_collection = typing.cast(Collection, other)
             return (
                 self._handle == other_collection._handle
-                and self._list_cache == other_collection._list_cache  # NOQA: W503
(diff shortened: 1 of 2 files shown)


nixnet/_session/collection.py
```diff
@@ -94,7 +94,7 @@ def __eq__(self, other):
             other_collection = typing.cast(Collection, other)
             return (
                 self._handle == other_collection._handle
-                and self._list_cache == other_collection._list_cache
+                and self._list_cache == other_collection._list_cache  # NOQA: W503
```
tkrebes, line 97: The code here actually matches our style guide. I would recommend ignoring W503 in the tox.ini in the flake8 section. In the future it would be good to use ni-python-styleguide for linting instead of flake8 directly, but for now if you run into warnings or errors that [ni-python-styleguide ignores](https://github.com/ni/python-styleguide/blob/be4560834ca09e9e1f08532481c841d91ba5cb57/ni_python_styleguide/config.ini#L31), you can just add them to tox.ini.

olsl21: Thanks for the info, I've added W503 to the flake8 ignore list in tox.ini in the latest commit.

tkrebes (approved):
Looks like the Python 3.6 run failed due to H903. Since Python 3.6 is EOL, I am going to ignore that error.

